PR #12, opened 2015-03-21
add shaders to recalbox

hstampfl:


digitalLumberjack:
Sorry i could not merge that pal, but with the new configgen it's now supported :)

PR #16, opened 2015-05-19
issues 188 & 183

rockaddicted:


digitalLumberjack:
Nice :)
Now we have to fix the issue when we have more than 22 systems...

PR #17, opened 2015-05-22
added new systems + modif simplesvgs theme

rockaddicted:
- added wonderswan libretro emulator
- added lutro libretro lua framework
- resized background gw/lynx simplesvgs from 1080p to 720p


digitalLumberjack:
ZUUUUUUBERBE

PR #40, opened 2015-11-04
Joysticks support improvement, including wiimote

nadenislamarre:
Summary of the 2 fixes :

1) make SDL2 considers any udev button
SDL2 ignores buttons which are not between BTN_MISC and KEY_MAX
RA_UDEV ignores buttons which are not between BTN_MISC and KEY_MAX or between KEY_UP and KEY_DOWN
2) make SDL2 assign buttons following the input.h file order as RA_UDEV does
an example joystick has 2 buttons : BTN_MISC and BTN_JOYSTICK

SDL2 will assign :
button 0 = BTN_JOYSTICK
button 1 = BTN_MISC

RA_UDEV will assign :
button 0 = BTN_MISC
button 1 = BTN_JOYSTICK

Buttons current behaviors :

SDL2:
between 0 and BTN_MISC-1 : ignored
between BTN_MISC and BTN_JOYSTICK-1 : second buttons group
between BTN_JOYSTICK and KEY_MAX : first buttons group

RA_UDEV:
between 0 and KEY_UP-1 : ignored
between KEY_UP and KEY_DOWN-1 : first buttons group
between KEY_DOWN and BTN_MISC-1 : ignored
between BTN_MISC and KEY_MAX : first buttons group

SDL2 updated to :
between 0 and KEY_MAX : first buttons group
no keys are ignored, so, potentially, we could update ra too if some poeple have buttons working in es but not in ra.
Risks :
1) SDL1 compatibiliy
theses modifications break compabitility between sdl1 and sdl2.
A joystick working with es but not working in ra could become not working on an other emulator based on sdl1.
2) Buttons reordering
A joystick configured in es could have to be reconfigured in case the buttons order has changed. However, if this occurs, it means that some buttons were already not working correctly in ra.

### DIFF
--- a/board/recalbox/fsoverlay/etc/udev/rules.d/99-wiimote.rules
+++ b/board/recalbox/fsoverlay/etc/udev/rules.d/99-wiimote.rules
@@ -1,2 +1,2 @@
 SUBSYSTEM=="input", ATTRS{name}=="Nintendo Wii Remote", MODE="0666", ENV{ID_INPUT_JOYSTICK}="1"
-
+SUBSYSTEM=="input", ATTRS{name}=="Nintendo Wii Remote Classic Controller", MODE="0666", ENV{ID_INPUT_JOYSTICK}="1"
--- a/board/recalbox/fsoverlay/etc/udev/rules.d/99-wiimote.rules
+++ b/board/recalbox/fsoverlay/etc/udev/rules.d/99-wiimote.rules
@@ -1,0 +1,2 @@
+SUBSYSTEM=="input", ATTRS{name}=="Nintendo Wii Remote", MODE="0666", ENV{ID_INPUT_JOYSTICK}="1"
+
--- a/board/recalbox/kernel_patches_3.19/linux-wiimote-abs-not-hat.patch
+++ b/board/recalbox/kernel_patches_3.19/linux-wiimote-abs-not-hat.patch
@@ -1,0 +1,54 @@
+--- a/drivers/hid/hid-wiimote-modules.c
++++ b/drivers/hid/hid-wiimote-modules.c
+1117,1122c1117,1122
+< 	input_report_abs(wdata->extension.input, ABS_HAT1X, lx - 0x20);
+< 	input_report_abs(wdata->extension.input, ABS_HAT1Y, ly - 0x20);
+< 	input_report_abs(wdata->extension.input, ABS_HAT2X, rx - 0x20);
+< 	input_report_abs(wdata->extension.input, ABS_HAT2Y, ry - 0x20);
+< 	input_report_abs(wdata->extension.input, ABS_HAT3X, rt);
+< 	input_report_abs(wdata->extension.input, ABS_HAT3Y, lt);
+---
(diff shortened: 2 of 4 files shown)


digitalLumberjack:
Hi dude,
Say you know i sometimes force the sdl2 driver in recalbox.conf, setting the sdl2 driver in retroarch. This is because some controllers have different mapping in udev and SDL2. 

Es get the SDL2 ones, and configgen map the retroarch settings with the SDL2 mapping. 
The main issue, is if i set SDL2 driver by default in retroarch, some controllers do not work anymore.

What could be the risk by merging your PR for this kind of controllers ?

nadenislamarre:
It will have no impact.
Because retroarch uses the sdl2 library which has the same modifications, so, it will map on the same button. So, i think that this sdl2 setting is no more required.
There is still one possible issue for what it would be required : i've not fixed retroarch to accept any button under BTN_MISC and not in KEY_UP, KEY_LEFT, ...
So, for example, a joystick driver sending a KEY_A event (letter A), would be button 1 in the new SDL, and ignored in retroarch => it would make the button not work in retroarch, but the main disadvantage it that it would create an offset between buttons of sdl and retroarch (in this case, force sdl2 would have a sence, but if that happend, i'll fix retroarch to be less restrictive)
Hum, i've a doubt, because the wiimote sends button A and B, and however, it works perfectly... i've to check i've not forced sdl2. (in the worse case, wiimote would not work without sdl2)

The problem could occurs only if you use a program based on sdl1 or the linux api joystick.
And in fact, these restriction comes from the aged linux api joystick and should not exist anymore in my humble opinion.

nadenislamarre:
For information, sdl2 uses udev,
it's not that they have a different mapping,
it's that SDL2 adds restriction for compatibility reasons with the old linux joystick api (such as a joystick is not supposed to send a-z keys, ...)

digitalLumberjack:
But it does have a different mapping depending on the driver in retroarch  for some controllers.

nadenislamarre:
Yes, but it will not be the case anymore.
I'll come on irc tonight to discuss, it will be easier.

The mapping was different before, because sdl2 has 2 set for mapping :
- first : numerate for BTN_1, BTN_2, ...
- seconds : numerate for the rest

example :
in input.h from linux, you've things like :

...
KEY_SOUTH 80
KEY_NORTH 81
...
BTN_1 100
BTN_2 101
...
KEY_YELLOW 110
KEY_BLUE 111
...

let's take a joystick having 3 buttons : KEY_NORTH, BTN_1 and KEY_BLUE.

Before my modifications,

in retroarch :
KEY_NORTH is button 1
BTN_1 is button 2
KEY_BLUE is button 3
because retroarch follows the input.h file order

in sdl2,
BTN_1 is button 1
KEY_NORTH is button 2
KEY_BLUE is button 3

because sdl2 first numerate buttons from BUTTON_1 to BUTTON_9, then the
others, for compatibility reason with the old linux joystick api.
I'm modified SDL2 to remove that point.
Now, sdl2 numerate exactly as retroarch does.

Nicolas

2015-11-18 13:45 GMT+01:00 digitalLumberjack notifications@github.com:

PR #41, opened 2015-11-07
Option in recalbox.conf to enable/disable samba at startup

nadenislamarre:

### DIFF
--- a/package/samba/S91smb
+++ b/package/samba/S91smb
@@ -15,6 +15,12 @@ mkdir -p /var/run/samba
 RETVAL=0
 
 start() {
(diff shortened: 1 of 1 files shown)


nadenislamarre:
Hum, don't close, my work is finally not perfect.
1) It doesn't respect the recalbox notation : system.samba = 1 instead of
system.samba.enabeld = 1
2) in the grep, it doesn't use the same syntax as in other /etc/init.d
services for example.

I've fix it tonight.

2015-11-18 13:06 GMT+01:00 digitalLumberjack notifications@github.com:

digitalLumberjack:
I'm on it i merged in order to modify it

PR #43, opened 2015-11-10
GCC 5.X fixups

igungor:
I've fixed 2 compilation problems that arose when I tried to build recalbox. 
Both of the patches are already in upstream or in a distro. Please test and apply.


digitalLumberjack:
Will be corrected with the buildroot update :)

igungor:
:+1:

PR #51, opened 2015-12-09
added new libretro cores

rockaddicted:
buildroot packages + buildroot configs
- amstrad cpc : libretro-cap32
- amiga : libretro-uae
- zx spectrum : libretro-fuse


digitalLumberjack:
That is great. Wich ones can start games directly ?

PR #56, opened 2015-12-10
/boot in ro in avoid crash issue

nadenislamarre:
noatime : performance (could be added in v3)
data=ordered : (avoid crash issue)

### DIFF
--- a/board/recalbox/fsoverlay/etc/fstab
+++ b/board/recalbox/fsoverlay/etc/fstab
@@ -9,6 +9,6 @@ tmpfs           /tmp           tmpfs    mode=1777         0      0
 tmpfs           /var           tmpfs    mode=1777         0      0
 tmpfs           /run           tmpfs    mode=1777         0      0
 sysfs		/sys	       sysfs    defaults	  0	 0
-/dev/mmcblk0p3	/recalbox/share ext4    defaults,rw	  0	 0
-/dev/mmcblk0p1	/boot		vfat	defaults,rw	  0	 0
+/dev/mmcblk0p3	/recalbox/share ext4    defaults,rw,noatime,data=ordered	  0	 0
+/dev/mmcblk0p1	/boot		vfat	defaults,ro	  0	 0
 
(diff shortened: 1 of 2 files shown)


substring:
Will need to remount /boot in rw anytime someone needs to change /boot/config.txt, and update wiki according to this

digitalLumberjack:
cool

pasnox:
What does noatime? Actually in QS i will rely a lot in file modified date
time for the rim scanner to only rescan file in change.

nadenislamarre:
http://lmgtfy.com/?q=noatime ;-)

Ne t'inquite pas, c'est juste sur les lectures de fichier (tu sais, l'info
dernier accès au fichier...) :

Linux has a special mount option for file systems called _noatime_ that can
be added to each line that addresses one file system in the /etc/fstab
file. If a file system has been mounted with this option, reading accesses
to the file system will no longer result in an update to the atime
information associated with the file like we have explained above. The
importance of the noatime setting is that it eliminates the need by the
system to make writes to the file system for files which are simply being
read. Since writes can be somewhat expensive, this can result in measurable
performance gains. Note that the write time information to a file will
continue to be updated anytime the file is written to.

2015-12-11 8:24 GMT+01:00 Nox P@sNox notifications@github.com:

substring:
Noatime means linux won't record thé ladt access time, not the last modification time. Si on impact for you

nadenislamarre:
et tant qu'à faire :

data=ordered, c'est le mode par défaut, mais bon.

data=ordered    (*) All data are forced directly out to the main file
            system prior to its metadata being committed to the
            journal.

in the default data=ordered
            mode, the data blocks of the new file are forced
            to disk before the rename() operation is
            committed.  This provides roughly the same level
            of guarantees as ext3, and avoids the
            "zero-length" problem that can happen when a
            system crashes before the delayed allocation
            blocks are forced to disk.

2015-12-11 8:46 GMT+01:00 Nicolas Adenis-Lamarre <
nicolas.adenis.lamarre@gmail.com>:

pasnox:
merci ;)

2015-12-11 8:48 GMT+01:00 Nicolas Adenis-Lamarre notifications@github.com:

PR #62, opened 2015-12-13
- Fixes SDL not checking values in the correct keymap when shift and …

nadenislamarre:
…alt are pressed.
- the patch doesn't include : english/french keyboard switching
- a bug noticed during the patch test : in v4, once the update message is displayed, the keyboard
  doesn't work anymore. If i start without the network wired, the keyboard works. If i go fast
  and test the keyboard before the message, it works. The issue occurs even when this patch is
  not applied. I don't know if it occurs in V3.

This patch could be applied in V3 too.

### DIFF
--- a/package/sdl2/sdl2-0002-shiftalt.patch
+++ b/package/sdl2/sdl2-0002-shiftalt.patch
@@ -1,0 +1,40 @@
+diff --git a/src/core/linux/SDL_evdev.c b/src/core/linux/SDL_evdev.c
+index 19333f2..97da3d7 100644
+--- a/src/core/linux/SDL_evdev.c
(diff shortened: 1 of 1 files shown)


digitalLumberjack:
- english/french keyboard switching -> does it means when i change the keyboard layout with `loadkeys` it won't work in ES ?
- The bug of the update screen occurs even without the patch ?

nadenislamarre:
i did ont try. but if you switch in fr kb it will work too.
yes. this is a différent bug.
Le 14 déc. 2015 10:03, "digitalLumberjack" notifications@github.com a
écrit :

digitalLumberjack:
Let's try that

PR #64, opened 2015-12-14
Update to 4.0.0

substring:
Update from main project


substring:
subs is a big baka ... wrong move

PR #70, opened 2015-12-17
recalbox.conf switch rpi1/rpi2 + updated recalbox-configgen package

rockaddicted:


nadenislamarre:
whoouhou !
is there a way to enable the multiplayer action to do on snes9x_next ?

digitalLumberjack:
~~Sorry pal it will not work. That copies the files in $target/recalbox/share but this directory is mounted with the share partition...~~

digitalLumberjack:
allright it does work i forgot i added this line : https://github.com/recalbox/recalbox-buildroot/blob/unified/board/recalbox/copy-recalbox-archives.sh#L20 sorry for the mistake

rockaddicted:
no problem dude.
About multi on snes9x_next, I think we have just to add this option 
`input_libretro_device_p2 = « 257 »` 
to this file : 
https://github.com/recalbox/recalbox-buildroot/blob/unified-4.0.0/board/recalbox/share/system/configs/retroarch/retroarchcustom.cfg.origin

PR #72, opened 2015-12-17
retroarch hotkey fix

digitalLumberjack:
uses the @kivutar patch to correct hotkey on select coin on arcade cores.


rockaddicted:
let's try this

digitalLumberjack:
@nadenislamarre the udev file has been modified since the last commit, is it ok with the patch ?

nadenislamarre:
sorry, i don't understand the question.
If it's about my retroarch patch for more allowed keys (same as sdl2), yes.

2015-12-17 20:03 GMT+01:00 digitalLumberjack notifications@github.com:

digitalLumberjack:
Yes it was about that :) thanks

PR #76, opened 2015-12-17
enabled by default multiplayer option on snes + added a udev rule

rockaddicted:


nadenislamarre:
hum,
i thought it was not enabled because of rpi1 of snes9x_next not by default...
So i am missing something : why it was not enabled before ?

rockaddicted:
Yes indeed, I thought that too.
I never tried this option before (I always only used pcengine and arcade to play with 3 others pals)
So, I enabled this option, and praticed some tests.
I tried the other systems, and snes with pocketsnes core, and no issues to play on single or multi.

In the retroarch input menu, I can see the "user 2 device type" switch from "retropad" to "unknown", but it seems that there is no impact on gameplay.

So I think we can test this on the alpha/beta 4.0.0, and rolling back if someone have an issue.

digitalLumberjack:
So great

PR #78, opened 2015-12-19
Buildroot: Rewrite post-image scripts.

pasnox:
Allow to use them from in source tree and out of source tree.
Also shrink the code to make it more readable.
Add support for own ssh key and sftp-server.
Fix boot and share .tar.xz permissions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 *._.DS_Store
 .config
 Thumbs.db
+authorized_keys
--- a/board/recalbox/copy-recalbox-archives.sh
+++ b/board/recalbox/copy-recalbox-archives.sh
@@ -1,26 +1,30 @@
-images="output/images"
+#!/bin/bash -e
 
-rm -rf $images/recalbox
-mkdir $images/recalbox
+# PWD = source dir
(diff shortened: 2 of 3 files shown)


digitalLumberjack:
@pasnox why not enable the sftp in buildroot ?

pasnox:
Dropbear does not comes with a sftp-server implementation - only openssh does.
We prefer Dropbear over openssh for lightweight memory footprint.
We can't install both dropbear and openssh at the same time from buildroot, so I need to build the sft-server on my own here and copy it.
It's only done when in 'dev' mode  - when u use a recalbox ssh key to avoid builds for standards release.

digitalLumberjack:
Sorry pal i don't like to have this in the production script for building target. Maybe we could create a package for the sftp server

pasnox:
Ok, I will update the merge request to change that.

pasnox:
Just rebased my changes after having removed ssh / sftp.
I still let the authorized_keys permissions fix because it's needed to avoid troubles, but no more ssh key is inserted.

digitalLumberjack:
Vendu !

PR #80, opened 2015-12-19
rolling back to official scummvm version

rockaddicted:
- official scummvm version is now using SDL2
- joystick support is now ok, but mouse sentivity with joystick is not excellent
- added build options to activate virtual keyboard
- patched sdl-event to toggle the virtual keyboard with the joytick


digitalLumberjack:
cool :)

PR #86, opened 2015-12-22
added kernel dtd

digitalLumberjack:


nadenislamarre:
hum didnt you break things ?
shairplay disabled ? and others ;-(
Le 22 déc. 2015 11:17, "digitalLumberjack" notifications@github.com a
écrit :

pasnox:
This seems to be default ON packages which get removed by 'saveconfigs', I guess it's ok.
Don't need to enable already defaulted enabled packages ?

digitalLumberjack:
yes it's ok it's enabled i verified. It's savedefconfig that removed them because of the transitive dependency

nadenislamarre:
OK good.
that was just in case.
thanks.
no computer to check myself.
Le 22 déc. 2015 12:14, "digitalLumberjack" notifications@github.com a
écrit :

PR #94, opened 2015-12-27
Unified 4.0.0 initshare merged nadenislamarre

nadenislamarre:
The script recalbox-patch-target.sh now fails once merged,
but i don't understand what the authorized keys files are supposed to do here.
I have not merged (so that it doesnt' fail), otherwise i would have just removed it, but perhaps it is a good reason to be here.

This patch allows to :
- separate user and system files (on 2 partitions)
- allow to user a usb key as user data directory with only one configuration which will be added to es menu in an other patch.

it has impact on samba shares.
share.tar.xz doesn't exist anymore : you can create an empty partition, and it will be used by default (INTERNAL mode)
You can create only 2 partitions now (/boot and /) and it will works (you've to switch configuration to an usbkey while it will swtich to a memory only user data directory).

You can now reinstall recalbox on a sdcard and find your data back while you've just to plug you're external key.
On the usb key, the directories are created automatically under the recalbox directory to not pollute the usb key.


digitalLumberjack:
Good job, i'll build and test that and we'll talk about it.

substring:
i'm impatiently waiting for the next build !

substring:
could you post your partitions.json and your partition_setup.sh ?
I removed data about the 3rd partition, but once recalbox is up, /recalbox/share didn't take the space left on the SD card

PR #107, opened 2016-01-03
Boot: remove some unusefull things

nadenislamarre:
S02splash : the program ends after 1 sec, no need to wait and kill
S33splashreboot : no nice to display shutdown screen at the startup process ;-) (and moreover, it contains sleep slowing down the boot process)
recalbox-patch-target.sh/S50kodi : remove S50kodi
bluetoothcontrollers.sh : keep the process open even if no device if found : if you plug a device, no need to reboot.


board/recalbox/fsoverlay/etc/init.d/S02splash
```diff
@@ -3,8 +3,6 @@
 case "$1" in
   start)
       fbv -f -i /recalbox/system/resources/splash/logo-loading.png &
-      sleep 1
```
digitalLumberjack, line 6: fbv doesn't stay alive if we do not killall ?

nadenislamarre, line 6: No, it doesn't.
It's not documented, but i can read the code for a confirmation.

2016-01-04 10:50 GMT+01:00 digitalLumberjack notifications@github.com:

digitalLumberjack:
Nice :)

PR #109, opened 2016-01-05
Boot sequence : faster / display version

nadenislamarre:
- S29 from 4s to 1.5s (parallelize python calls to get parameter values) ; S31es is called at t+5 now (instead of t+7)
- splash : reduce the size + display the version number on left
- remove unused .template old files


digitalLumberjack:
Really good work pal.

nadenislamarre:
which one ?
https://fr.wikipedia.org/wiki/Pal

PR #110, opened 2016-01-07
filesystem: create the internal partition if it doesn't exist (fat32)

nadenislamarre:


digitalLumberjack:
Nice, but that will not work with noobs. Not a problem,  think we might create it directly from noobs as it creates a tricky partition table with extended partitions etc...

pasnox:
Any reason to rely on parted while fdisk is already available?

nadenislamarre:
You're right.
I didn't see fdisk was available.
However, i feel fdisk is more for enduser while parted is more for scripting.
I find parted more robust (i can tell exactly was to do), but if you find the fdisk equivalent calls, we could disable parted and that would be very nice ;-)

parted -s /dev/mmcblk0 -m unit b mkpart primary "$PSTART" 100%

versus

echo "(${rootKSize}/1024)/1" | bc`
(
    echo -e "o" ; # Erase partition table

```
echo -e "n" ; # New partition
echo -e "p" ; # Primary partition
echo -e "1" ; # Partition 1 (Boot)
echo -e ""  ; # default - Start after preceding partition
echo -e "+${bootKSize}K" ; # Boot partition size
```

) | fdisk /dev/mmcblk0p3

pasnox:
Indeed fdisk is a bit more verbose. I guess it depend how much space is
needed by parted to make the choice.

PR #111, opened 2016-01-07
sd : update the script to take only 2 partitions

nadenislamarre:
need to be updated to make more tests, but works.
make to boot a bit bigger in case of update.

### DIFF
--- a/scripts/linux/mk-sdcard-img.sh
+++ b/scripts/linux/mk-sdcard-img.sh
@@ -10,12 +10,12 @@ ScriptDir="`dirname ${0}`"
 
 # Return uncompressed size for the given tar.xz file
 # $1: XZ file path
-function getUncompressedFileSize() {
+getUncompressedFileSize() {
     local size=`xz --robot --list "${1}" | grep totals | cut -d $'\t' -f5,5`
     local extra=$((1024*1024*10))
     local fat32_minimum=33548800
     local size=$((${size}+${extra}))
-    local size=`echo "scale=2;${size}*1.2" | bc`
+    local size=$((${size}*14/10))
 
     if [ "`basename \"${1}\"`" == "boot.tar.xz" ]; then
         if [ "`echo \"${size}/1\" | bc`" -lt "${fat32_minimum}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/linux/mk-sdcard-img.sh
```diff
@@ -88,8 +85,7 @@ for image in ${images}; do
     echo "* Cleaning ${image}..."
     umount "${SDCARD_BOOT_PWD}" > /dev/null || true
     umount "${SDCARD_ROOT_PWD}" > /dev/null || true
-    umount "${SDCARD_SHARE_PWD}" > /dev/null || true
-    kpartx -sd "${image}" > /dev/null || true
+    kpartx -sd "${image}" > /dev/null
```
pasnox, line 88: Why removing dev null here? Sometime the img exists but is not mounted at all

scripts/linux/mk-sdcard-img.sh
```diff
@@ -88,8 +85,7 @@ for image in ${images}; do
...
     rm -f "${image}" > /dev/null
```
pasnox, line 89: I mean || true

nadenislamarre:
Hi,

i'll do some other changes on the script, but it works already very well ;-)
I suggest i finish and we discuss on irc about the changes ?

scripts/linux/mk-sdcard-img.sh
pasnox, line 88: I mean || true

scripts/linux/mk-sdcard-img.sh
```diff
@@ -156,68 +142,43 @@ do
...
-fi
-releaseUser="`basename \"${releaseHome}\"`"
-releaseFilePath="${releaseHome}/public_html/${baseName}.tar.xz"
-mkdir -p "`dirname \"${releaseFilePath}\"`"
-chown -R ${releaseUser}:www-data "${SDCARD_IMG_FILE_PATH}"
-chmod -R u+rwx,g+rwx,o+rx "${SDCARD_IMG_FILE_PATH}"
-tar -cJvf "${releaseFilePath}" -C "`dirname \"${SDCARD_IMG_FILE_PATH}\"`" "`basename \"${SDCARD_IMG_FILE_PATH}\"`" > /dev/null
-chown -R ${releaseUser}:www-data "${releaseFilePath}"
-chmod -R u+rwx,g+rwx,o+rx "${releaseFilePath}"
-rm "${SDCARD_IMG_FILE_PATH}"
```
pasnox, line 220: Move that to an unused function instead of removing it - i do want to use it on my server

pasnox:
Ok. ;)

nadenislamarre:
the "Moving to testing area..." seemed to be too specific to your needs (and not related to the sd card creating).
but we can move it to an other script calling the sd card creating and then moving the img to your www.
Please note that /home/recalbox, public_html, www-data are all specifics and probably unwanted for most people. This part is very specific to your needs i guess for the way you deliver the sd card.
However, very good job. It should be used for nightly + official builds in my opinion.

digitalLumberjack:
It will be used in nightly buids soon, but i agree with you the Moving is specific to the build environnement, i will remove that and use it in a specific script on my server.

nadenislamarre:
I will add options in the script,
like the sd card file path, the sd compression and the help.

nadenislamarre:
Are we sure we need openjdk-8-jdk ?

PR #112, opened 2016-01-08
scripts : update mk-sdcard

nadenislamarre:

### DIFF
--- a/scripts/linux/install-ubuntu-deps.sh
+++ b/scripts/linux/install-ubuntu-deps.sh
@@ -17,5 +17,4 @@ apt-get install \
     texinfo \
     zip \
     openjdk-8-jdk \
-    kpartx \
     inotify-tools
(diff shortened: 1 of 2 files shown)


pasnox:
Nested "if" looks a bit unreadable in my opinion.
Can t understand the need of double quotes in strings?
Why changed compression from tar.xz to tar.gz?
The dd-sdcard.sh script would require changes too.

nadenislamarre:
pasnox : hehe, we could talk on irc ;-)
- hum, what do you suggest instead of nested "if" ? bash -e didn't allow to correctly handle errors (for example umount if mount success but the following command fails)
- for the quotes, you must do it for strings to handle spaces or empty strings, but perhaps i did it sometimes for number which is not really a problem
- the compression gz is only used for the img, not for the tar. However, seeing your 2nd script, i guess i've mistakenly not seen that you compressed already the image.

i removed kpartx because i far i understand, kpart -d doesn't remove exaclty all what kpart -a does.
using losetup was not more complicated in fact.
I hesitated to remove . "${ScriptDir}/br-shared.sh" ; should i ? i i don't see in which case it can be usefull.

nadenislamarre:
hum, i will replace the xz/gzip by a zip for windows users...

PR #114, opened 2016-01-10
Unified 4.0.0 upgrade nadenislamarre

nadenislamarre:
Upgrade at startup if files (root.tar.xz and boot.tar.xz) found in /recalbox/share/system/upgrade


digitalLumberjack:
Huge ! I'll update the build script for the recalbox.version :) Very good job pal (the buddy pal)

PR #118, opened 2016-01-12
Unified 4.0.0 ntfs nadenislamarre

nadenislamarre:


nadenislamarre:
Hum, a ntfsfix is welcome in case the device in not writable in case of bad device removal (like in S11).
I'll add this later.
For the moment, if the device is badly removed, until it's plugged on a windows or ntfsfix is run by a linux, it will be remounted in ro.

PR #119, opened 2016-01-13
kodi : add backward/forward while playing from joystick

nadenislamarre:


digitalLumberjack:
That's nice :)

PR #125, opened 2016-01-17
kodi : enable lirc in kodi 15

nadenislamarre:
in kodi 15, lirc is enabled by default.
    --disable-lirc is used to disable lirc.
    Unfortunatly, --*-lirc does the same, including --enabled-lirc.


digitalLumberjack:
Nice !

PR #138, opened 2016-01-22
storage: back to ext4 as default for internal

nadenislamarre:
internal partitions currently formatted in fat32 will continue to work after the upgrade,
but without the flush option,
and kodi will not be able to install some plugins (utf8 missing).
I advice to rebuild the .img while we will stay definitively in ext4 for the internal while it's really more robust
than
fat => no journal
ntfs => undocumented protocol ; not completly supported on linux (on journal features)


digitalLumberjack:
It think it's the best solution.

PR #148, opened 2016-01-25
mount : only 1 script to mount to control all mounting options

nadenislamarre:
1) only one place to mount for : S11share (mount share), usbmount and the future recalbox-sync
2) on ntfs : check to write a file, or run ntfsfix
3) on vfat : run fsck except for usbmount via a script option if writing a file fails


digitalLumberjack:
Good idea :)

PR #151, opened 2016-01-25
take only one parameter for recalbox-config.sh storage

digitalLumberjack:
@nadenislamarre does it matter for you ?

### DIFF
--- a/board/recalbox/fsoverlay/recalbox/share_init/system/.emulationstation/es_settings.cfg
+++ b/board/recalbox/fsoverlay/recalbox/share_init/system/.emulationstation/es_settings.cfg
@@ -14,5 +14,3 @@
 <string name="ThemeSet" value="simplelight" />
 <string name="TransitionStyle" value="fade" />
 <bool name="Overscan" value="false" />
-<string name="Overclock" value="none" />
-<string name="Overclock-rpi2" value="none-rpi2" />


digitalLumberjack:
just seen the supplementary "" in the echo, i did not test the thing, maybe we could just remove the DEV from the command and from the list and add it when echoing in the /boot/recalbox.conf
All of that to keep ES stupid for this.

digitalLumberjack:
Works like that, if you have a better solution i take it, but right now it makes it fully working from ES.

nadenislamarre:
no, that's ok for only 1 parameter.
By the way, only the name (and not the uuid) should be displayed in the
interface.

2016-01-26 8:19 GMT+01:00 digitalLumberjack notifications@github.com:

nadenislamarre:
you can do this if you prefer :
recalbox-config.sh storage list
recalbox-config.sh storage current
recalbox-config.sh storage INTERNAL
recalbox-config.sh storage RAM
recalbox-config.sh storage ANYEXTERNAL
recalbox-config.sh storage [UUID]

RAM should not be displayed in the interface while it's usefull only for tests.

digitalLumberjack:
If fact the current code work (with the last commit)
The only thing is that is displays DEV UIDNONAME
Maybe we could display UID NONAME (to differ 2 NONAME) in ES

nadenislamarre:
it's not needed to display 2 noname i think while the uuid is not helpfull
at all for the user.
Having 2 noname or 2 noname uuid is the same for the user.
Most of usb keys have a name.
People will not plug several usb key i guess too.

PR #152, opened 2016-01-25
sync : synchronize your current share with an other fs

nadenislamarre:
(usb key, or internal if running already on an usb key)

to backup/copy to a friend,
1) plug a usb key
type : recalbox-sync.sh list
then recalbox-sync.sh sync uuid_of_the_key

2) to restore on the internal device
use a usb key as the share
recalbox-sync.sh sync INTERNAL

something to integrate similarly to share boot on es.
note that this script can be destructive : it synchronize the shares, so if one rom is not available on the source, it will be removed on the target.

More you run this script, faster it is.
Note that on ntfs, it is slower (than on fat, and really slower than on ext4).


digitalLumberjack:
Nice !

PR #155, opened 2016-01-27
upgrade : implement the stable/unstable/nightly system

nadenislamarre:


package/recalbox-system/recalbox-system.mk
```diff
@@ -9,21 +9,13 @@ RECALBOX_SYSTEM_SOURCE=
 RECALBOX_SYSTEM_VERSION=0.2
 
 ifeq ($(BR2_cortex_a7),y)
-	RECALBOX_SYSTEM_RPI_VERSION=rpi2-unstable
+	RECALBOX_SYSTEM_RPI_VERSION=rpi2
```
digitalLumberjack, line 12: maybe we could use only one variable ?

nadenislamarre, line 12: maybe ;-)
i know you would change that because i guess that the BR_FINAL can be
removed too.
So, do as you wish ;-)

please note that the wiki is to upgrade because i used the name you gave me
for recalbox.conf + the wiki is missing the /rpi1 or /rpi2 path.

Nicolas

2016-01-28 8:18 GMT+01:00 digitalLumberjack notifications@github.com:

nadenislamarre:
> > add the values at the end of the line like others : (stable, unstable, nightly)
> > we must not use things at the end because a nightly will become a future realease, without recompiling.

substring:
Dunno if you both had a talk about it, but i find the "Stable - Beta - Dev" labels much more explicit. That setting will prolly be available to end users. The "'stable/unstable/nightly" is too much "developer-oriented", whereas Beta and development are a way more common among "users"

digitalLumberjack:
soon.

digitalLumberjack:
I close as it's ok in rb-4.0.X

paradadf:
Hi! Right now you have "stable - beta - unstable". Shouldn't it better be as @substring said: "stable - beta - DEV" instead?

Also, I don't know if you noticed, but everytime after unstable is chosen and the system is updated, unstable is changed to stable.

If both are wanted features, then just ignore my comment.

digitalLumberjack:
> Right now you have "stable - beta - unstable".  

cannot change too much risk :)

> but everytime after unstable is chosen and the system is updated, unstable is changed to stable.

not normal, the recalbox.conf is not overwritten when you update the system. Are you sure about that recalbox.conf updates.type is changed ??

paradadf:
I am pretty sure, but let me confirm it to you tomorrow after my next update ;)

paradadf:
Nothing changed this time. I don't know what happened. Sometimes the changes I do on recalbox.conf over the browser aren't saved.

About the the "stable - beta - dev" thing... you should consider that people that just updated their systems from 4.0.0 beta1 actually have "stable - unstable - nightly" and without updating manually the recalbox.conf, they have something outdated.

substring:
As a wider consideration, updating the recalbox.conf is definitely a problem ...

PR #157, opened 2016-01-28
vfat : bluetooth support

nadenislamarre:


digitalLumberjack:
Good idea :)

PR #167, opened 2016-02-02
joystick: handle plug/unplug events when no windowing system already …

nadenislamarre:
…does it

Should fix #614


digitalLumberjack:
That's cool :)

PR #175, opened 2016-02-08
Rajout de nouveaux core pour recalbox 4.0.0

ghost:


nadenislamarre:
Nice. The 5 are working correctly with rpi1 and rpi2 or just rpi2 (can somebody test) ?
The directory for dosbox seems missing in share_init/roms and in es_system ?
roms must be at least of the form : .exe .cmd and .com i guess.

nadenislamarre:
only in es_system. The directory exists.
Do we really want links on http://emulparadize ?

substring:
the keyboard doesn't work yet in dosbox
no EP links, although some other readme are qquite ambiguous too
the main config.in is not modified so we can select packages through menuconfig
no defconfig for rpi1 and rpi2
I'll give a try to the libretro core of dosbox and see if it works better than compiling straight from the source

DrNosh:
For dosbox, keyboard and sdl video driver don't work but the dosbox libretro emulator work. In the nearly futur add keyboard config and conffigen option for pokemini.

digitalLumberjack:
Closing this one as the other PR split the features

PR #177, opened 2016-02-09
updated theme simplelight

rockaddicted:
Added missing themes, but without background musics (seen there will have a workshop on this)


digitalLumberjack:
:+1:

PR #179, opened 2016-02-09
Unified 4.0.0 update recalboxconf n64

rockaddicted:
Removed N64 section in rpi1's recalbox.conf and updated rpi2's one.


nadenislamarre:
digit,
is it possible to make a zip and put it on the server on a section :
/extra/themes

tell me once done to do :
i suggest a script :
recalbox-extra.sh

so that recalbox-extra.sh zoid download and extra the theme on the usb key at the correct place.

digitalLumberjack:
Yes really good idea. We can also add a script that would get the recalbox-gamelists project resources and put it in the good directories :)

PR #194, opened 2016-02-17
pm2 daemon should be started in the API directory

neolao:


digitalLumberjack:
Isn't it dangerous to change the current directory in a init.d script (just asking i don't think it is)

neolao:
I will try to find another way to run the command.

neolao:
Ok I found something. I set the environment variable API_HOME in the script.
The last release of the API uses it.

digitalLumberjack:
Thank you :) That is the best solution. I checked and i saw that cd is ok in init.d scripts as each script is started in a new bash (that seems normal :)

PR #196, opened 2016-02-19
Noobs adaptation + new update urls

digitalLumberjack:
So 
- now we have the partition for internal storage directly in config.txt.
- it's by default /dev/mmcblk0p3 for img install and it's dynamically patched by the noobs installation : https://github.com/recalbox/recalbox-rescue/commit/50d28c85518a566d2dd84dda47c1ef97b3c0acc8
- the update url is now archive.recalbox.com/4/
- in the url, last link  is now in the differents arch directory, not in root dir

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S11share
+++ b/board/recalbox/fsoverlay/etc/init.d/S11share
@@ -11,6 +11,7 @@
 
 SHARECONFFILE="/boot/recalbox.conf"
 INTERNALDEVICE=`cat /boot/config.txt | grep "internal=" | head -n1 | cut -d'=' -f2`
+[[ "$?" -ne "0" || "$INTERNALDEVICE" == "" ]] && INTERNALDEVICE=/dev/mmcblk0p3
 INTERNALDEVICETYPE="ext4" # faster than waiting udev to initialize to get the type
 MAXTRY=5
 NTRY=0
--- a/board/recalbox/fsoverlay/etc/init.d/S11share
+++ b/board/recalbox/fsoverlay/etc/init.d/S11share
@@ -10,8 +10,11 @@
 ###
 
 SHARECONFFILE="/boot/recalbox.conf"
-INTERNALDEVICE=`cat /boot/config.txt | grep "internal=" | head -n1 | cut -d'=' -f2`
+INTERNALDEVICE=`cat ${SHARECONFFILE} | grep "internal=" | head -n1 | cut -d'=' -f2`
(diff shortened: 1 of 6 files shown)


nadenislamarre:
In case internal= doesnt exist in config.txt, a default value is missing.
Please use the  /boot/recalbox.txt containing the share mount for specific recalbox values.

digitalLumberjack:
the way  /boot/recalbox.txt is used makes it overwritten by the recalbox-config.sh
We must have a value that is still.

We could use an other file if you think it's better, but now i changed the config.txt in build + patched it in noobs, so we cannot have an empty value, except if the user remove that.

nadenislamarre:
Yes, we should change the way recalbox.txt works or merge with config.txt
to not have 2 ways to do configuration.
Le 20 févr. 2016 10:11, "digitalLumberjack" notifications@github.com a
écrit :

digitalLumberjack:
What do you think ? everything in config.txt ? internal= and sharedevice=

nadenislamarre:
a custom recalbox.txt looks good. but all in config.txt is ok too.
Le 20 févr. 2016 15:20, "digitalLumberjack" notifications@github.com a
écrit :

LaurentMarchelli:
I expect an issue with this, before assuming internal device is the /dev/mmcblk0p3, /etc/fstab must be checked before.
In fact, if the configuration is made with Noobs and 3 partitions, as it is configured in  recalbox-rescue, with NOOBS 1.7 the "share" partition will not be the /dev/mmcblk0p3, but the /dev/mmcblk0p8 or /dev/mmcblk0p10 and this information will be available from /etc/fstab.

digitalLumberjack:
Check this : https://github.com/recalbox/recalbox-rescue/blob/rb-4/sdcontent/os/recalboxOS-rpi2/partition_setup.sh

Nothing in fstab, no issues :)

LaurentMarchelli:
Right for new install :-)
But, for those who will upgrade from previous version ?

LaurentMarchelli:
May I use the same configuration file (/boot/recalbox-boot.conf) to store external switch configuration ?

By external switch configuration I mean :
- RemotePi Board 2013 (http://www.msldigital.com)
- RemotePi Board 2015 (http://www.msldigital.com)
- Mausberry Circuits (http://mausberry-circuits.myshopify.com)
- ATXRaspi (http://lowpowerlab.com/atxraspi/)
  ...

Remote Pi scripts #216 (Open)
https://github.com/recalbox/recalbox-os/issues/216

digitalLumberjack:
Yes you can

LaurentMarchelli:
Ok thanks.

PR #204, opened 2016-02-24
Default lircd.conf could manage default remote control devices.

LaurentMarchelli:
Reuse of OpenELEC LIRC configuration file adaption for Raspberry Pi & Kodi
https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/sysutils/lirc/config/lircd.conf.rpi

Original LIRC conf files can be found at :
http://lirc-remotes.sourceforge.net/remotes-table.html

I'm using Harmony One with Microsoft Media Center Edition with Rasplex (OpenELEC) and it works fine with it.

### DIFF
--- a/board/recalbox/fsoverlay/recalbox/scripts/recalbox-config.sh
+++ b/board/recalbox/fsoverlay/recalbox/scripts/recalbox-config.sh
@@ -9,7 +9,7 @@ command="$1"
 mode="$2"
 extra1="$3"
 extra2="$4"
-version=`cat /recalbox/recalbox.arch`
+arch=`cat /recalbox/recalbox.arch`
 
 recalboxupdateurl="http://archive.recalbox.com/4"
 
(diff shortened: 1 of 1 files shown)


digitalLumberjack:
Thanks pal, i'll wait for @nadenislamarre go to merge, he is in charge of lirc

nadenislamarre:
hum, i guess it works.
however, for people having not one of these remotes, it means that we have to clear the file and lost the information.

I suggest
1) include your changes
2) add an include in  /board/recalbox/fsoverlay/recalbox/share_init/system/.config/lirc/lircd.conf
of /board/recalbox/fsoverlay/recalbox/share/system/.config/lirc/lircd.conf

3) remove the copy of the lircd.conf in S12 from share_init to share
4) in S25, use the share_init config file (which include the share config)
5) update the wiki, but i guess there is probably nothing to do

LaurentMarchelli : thanks for your help. Do you feel to do steps 1 to 5 of if you prefer, i can do them.

LaurentMarchelli:
I think it could be a bit different, from a process point of view, I'm not  really in favor of including user specific file inside system generic file, so I propose :
1. Rename the modified file : /recalbox/share_init/system/.config/lirc/lircd.conf.rpi
2. Change /recalbox/share_init/system/.config/lircd.conf
`# THIS FILE MUST BE CONFIGURED TO MAP YOUR REMOTE`
`# SEE https://github.com/digitalLumberjack/recalbox-os/wiki`
`include "/recalbox/share_init/system/.config/lirc/lircd.conf.rpi"`
3. Update the wiki :-)
New installation will take the benefit, for updaters we need anyway, to find a smart process to update files located into /recalbox/share based on news parameters added in /recalbox/share_init files, by the update.

LaurentMarchelli:
I have reverted my commit, to change it.
But ... include does not seems to work with lirc, any idea ?

nadenislamarre:
check the file /etc/lirc/lircd.conf under your recalbox, it contains an
include.

LaurentMarchelli:
The real answer seems to be : 
Lirc does not manage include with absolute path. :-)
This doesn't work :
include "/recalbox/share/system/.config/lirc/lircd.rpi.conf"
But this works :
include "../../../../share/system/.config/lirc/lircd.rpi.conf"

PR #205, opened 2016-02-25
Fix recalbox/recalbox-os#216

LaurentMarchelli:
Add external switch support, by external switch configuration I mean :
- ATXRaspi (http://lowpowerlab.com/atxraspi/)
- Mausberry Circuits (http://mausberry-circuits.myshopify.com)
- RemotePi Board 2013 (http://www.msldigital.com)
- RemotePi Board 2015 (http://www.msldigital.com)

Remote Pi scripts #216 (Open)
https://github.com/recalbox/recalbox-os/issues/216

Tested on RemotePi Board 2015, works when :
- Switch off is called by the button
- Shutdown is called by EmulationStation.
- ATXRaspi & Mausberry may need improvement to reset hardware button status.


digitalLumberjack:
Hi, nice pr. I know now why you asked me if you could use recalbox-boot.conf to store informations.

Let's just see if it worth letting that in this file, or putting the switch in recalbox.conf.
What would be wrong with set a system.power.switch=xxxxx in recalbox.conf.

That would mean several things. If we store the config on an external storage the configuration will be lost when we remove it. But this is already the case for the GPIO switch for example, wich is dependant of the hardware too.

SO maybe we can put everything in recalbox.conf, letting people know that is it stored in that (that also would allow a simpler configuration, all in one file) or we put all hardware dependant switches in recalbox-boot.conf.

Second thing, do you think you could add a raspberrypi pin 5/6 switch ?

LaurentMarchelli:
I agree, it's better to have all configurations parameters in the same file.
Even if, boot seems more in line with the Raspberry Pi config.txt file, to store hardware specific parameters, I guess there are more benefits to keep it on share :
- boot is read-only, to modify the file, /boot need to be remounted rw
- share is read-write
- share is available and can be modified with samba.
- keeping on share is less modifications. :-)
  So, I can move my parameter to system.power.switch-xxx into recalbox.conf for both RP1 and RP2

For the second question, no problem, it is designed to be extended :
- Give me a fancy name that support regex "\w*"
- Give me a link where I can get the information on which GPIO I should address.

digitalLumberjack:
Hi, 
so you can find informations here : http://raspberrypi.stackexchange.com/questions/13203/creating-halt-wake-button
It's the pins 5/6 that would have a switch button : the script listen for an edge, then shutdown -h now. When the rpi is off, an other edge will boot it.

For the names in recalbox.conf : we already provide names with dots, i think it's better to stay in this mode.

 system.power.switch may be the good name.

LaurentMarchelli:
I got it, done with system.power.switch at the beginning of section A.
I will have a look, did u find a pretty name for the 5/6 switch ? :-)

digitalLumberjack:
Really nice :) Thank you.
Maybe there is one more thing we can add for the raspberry pi wakehalt : 
The choice beween a on/off button with state, or a temporary button with no state.

What do you think about that ?

LaurentMarchelli:
What you mean by state ? 
Does it means, select another GPIO for led state ? : 
- Set it to 1 to switch on the led when the RPI is awake.
- Set it to 0 to switch off the led when the RPI is halt.
- And eventually, alternate state during the shutdown to have a blinking led ?

If it is what u want, it is not a big deal, I will add a second parameter to the line for led management :
system.power.switch=RPI_HALTWAKE,5,6

digitalLumberjack:
No that is more like the difference between : 
Momentary button VS on/off switch (sorry i did not have the words)
So the code you did is made for a on/off switch. But a momentary button will act differently : the button is always open (so the gpio state is up) and we must wait for an edge (a press on the button will set the state to 0)

In python it's like that : 
GPIO.wait_for_edge(0, GPIO.FALLING)

But we don't have python gpio support. maybe it's time to add that. I'll take a look this weekend.

digitalLumberjack:
Note to myself : add a wiki about that, and let people know that they have to put a 10kohms if it's an on/off switch

LaurentMarchelli:
Do u mean I implemented a push button (push event) and you wanted a power switch ON/OFF ?
Or is it a problem between low and high (0 /1) ?

LaurentMarchelli:
Do u have a schema of electronic implementation, I guess it will help.

digitalLumberjack:
So i must go now i don't have time to check for schema, i will try to show you later this day.

What we need is both implementations : 
- The momentary button : like an keyboard button
  - when the pi is down, it wakes it up as you make an edge between the two pins
  - when the pi is up, the gpio is always in state 1(3.3v) as the contact is not made with the ground, so we must wait for the falling of the voltage that would mean someone press the button. This edge will be very quick because it's a simple press (the 1 second sleep will never see it i think)
- The on/off swtich : like the supernes on off switch
  - when the pi is down, it wakes it up as you make an edge between the two pins
  - when the pi is up, the gpio is always in state 0 (0v) as the contact with the pin 6 gnd is always made, so  we must wait for a rising of the voltage that would mean someone turn the switch to its off state.

Sorry i reach my english limit when it's about technical vocabulary like that, i may not be clear enough.

LaurentMarchelli:
Should we deliver existing switch like it is and we'll create another pr for your system improvement ?
I understood you wanted to detect state modification for one to another instead than waiting for an expecting state, even if, it could be done in the current code, we still have the issue of the polling delay (1s)

digitalLumberjack:
Ok so let's just talk about the switch you implemented. It's an on/off switch, but your are checking that the gpio state is !=0 but it will be 0 if we are considering that it's connected to the ground (as it is like that we wake up the pi)
And we must enable the internal pullup to avoid short circuit. I know how to do it in c, python, even assembly, but not in bash :(

If you still want to do that that is great, but don't hesitate to tell me to do it if it's not the case ;)

LaurentMarchelli:
I guess there is an mis-understanding, I really apologize for this.
At the beginning, I only wanted to implement the Remote Pi 2005 for my personal needs, but as most of the system worked the same, I though it was a good idea to have something generic.
Then, I tried to implement, what you requested for with the information I had.
Sorry, if I misunderstood your needs, if you have the C example, I would try to convert it for you.

digitalLumberjack:
Hum who misunderstood who ? :)
I was trying to tell you that i was sorry to "impose" you this work after you did a really good pr, and if you did not want to do it i would really understand, i was never thinking of criticizing your effort :)

You do not have to be sorry for anything :P

digitalLumberjack:
But i checked and it seems not possible to set internal pullups/down with bash only. So as my plan was to integrate python gpio lib that allow the wait_for_edge and internal pullups/down, maybe we can wait for that to integrate the rpi 5/6 wake feature. 
https://github.com/recalbox/recalbox-os/issues/299

LaurentMarchelli:
No problem, do not worry, I didn't get it as a criticism, I understood that you also wanted to go forward. ;-)
I proposed a global architecture to support the powerswitch system, and the 5/6 wake feature could be part of this architecture, it was my point of view.
To complete it, if bash is not enough, we could change the raspberry_start() content, calling an external blocking program that does the expected work.
Even if the 5/6 is not finished yet, we have an entry point for the complete the implementation, should we block this PR, with 4 features, for an incomplete one more ?
If we publish it, maybe someone will find a smart way to do it, and maybe someone will also propose the gui to configure it. It is the way collaborative work works, isn't it ? 
What do u think about it ? :-)

digitalLumberjack:
Yes we will merge that :)
Can you remove the 5/6 pin code, so it's fully functional with all the boards. ? I'll merge then.

LaurentMarchelli:
Done, code is clean.
Thank's :-)

Laddyacris:
Hello 
feedback into this post for REMOTEPIBOARD_2005 : http://blog.recalbox.com/forums/topic/issue-216/#post-25239

LaurentMarchelli:
thx for the feedback :-)

Laddyacris:
Hello LaurentMarchelli
on our forum , Lukas asks : 

>  Is there any chance to get support for the “Witty Pi”-Board (see [here](http://www.uugear.com/witty-pi-realtime-clock-power-management-for-raspberry-pi/)) with 4.0.0?
> 
> Being able to shutdown using the board would be completely sufficiant for me, no need for RTC and timers.

LaurentMarchelli:
I gonna have a look this we.
Regards
Laurent

PR #213, opened 2016-02-28
kodilauncher: kill kodi if it fails to stop. Reboot/Shutdown if wanted

nadenislamarre:


digitalLumberjack:
Bonne idée le coup du fichier de logs :)

PR #218, opened 2016-02-29
Update 8bitdo FC30 Joystick Bluetooth

neolao:
**Warning /!\**
The firmware 1.41 does not work.
I switched to 1.40 to map the gamepad.

### DIFF
--- a/board/recalbox/fsoverlay/recalbox/share_init/system/.emulationstation/es_input.cfg
+++ b/board/recalbox/fsoverlay/recalbox/share_init/system/.emulationstation/es_input.cfg
@@ -582,24 +582,27 @@
         <input name="x" type="button" id="4" value="1" code="308" />
         <input name="y" type="button" id="3" value="1" code="307" />
     </inputConfig>
-    <!-- 8Bitdo Joy Bluetooth -->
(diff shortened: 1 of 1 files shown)


digitalLumberjack:
Can you check if the firmware change the name of the controller when you try to syncrhonise the bluetooth ? (double check spaces etc)

neolao:
It's a weird behavior. The pairing can generate a joystick named like a PS3 controller. Here is the output of SDL :

```
$ /usr/recalbox-api/libs/bin/joysticks-linux-arm
```

```
Joystick 1 :
- JoyID = 1
- GUID = 060000004c0500006802000000010000
- Name = PLAYSTATION(R)3 Controller (64:03:D8:98:7F:06)
- Device path = /dev/input/js1
```

I think the firmware 1.41 (http://8bitdo.com/fc30joystick/Firmware.html) try to simulate a PS3 Controller by default.

When I try to map buttons, I lost connection after the first button.

digitalLumberjack:
Ra. Even with Start + R ?

neolao:
Ok it works now with the firmware 1.41, simply with `HOME`. I don't know why I was stuck on the PS3 mode.

PR #222, opened 2016-03-01
Add package RPi-GPIO #299

MikaXII:
With a `make python-rpigpio` the package is installed correctly in `usr/lib/python2.7/site-packages/`
But I can't try it.


digitalLumberjack:
Let's try that.
Thanks pal

ian57:
ok good. some user are waiting for that to use GPIO on their own

PR #228, opened 2016-03-03
Add dosbox 0.74

lmerckx:
Add **dosbox 0.74**:
- SVN version from SourceForge (0.74 version, last revision r3969)
- Patched to avoid keyboard problems

Needs following files/dependencies to work:
- libSDL-1.2.so.0.11.4 (patched SDL library - see other PR)
- libSDL_sound-1.0.so.1.0.2 (sdl_sound library - needed by dosbox)
- libSDL_sound-1.0.so.1 -> libSDL_sound-1.0.so.1.0.2 (symbolic link)
- libSDL_sound.so -> libSDL_sound-1.0.so.1.0.2 (symbolic link)
- recalbox-configgen PR to launch dosbox games


nadenislamarre:
Oh my god !
you rock !

digitalLumberjack:
Huge :)
Would it be envisageable to create the package for the patched versions of sdl ?

lmerckx:
Some additional notes:
- You need to add a new engine in es_systems.cfg:

```
  <system>
        <fullname>PC (x86)</fullname>
        <name>pc</name>
        <path>/recalbox/share/roms/pc</path>
        <extension>.pc .PC</extension>
        <command>python /usr/lib/python2.7/site-packages/configgen/emulatorlauncher.pyc %CONTROLLERSCONFIG% -system %SYSTEM% -rom %ROM% -emulator %EMULATOR% -core %CORE%</command>
        <platform>pc</platform>
        <theme>pc</theme>
    </system>
```
- Default configuration file is /recalbox/share/system/.dosbox/dosbox-0.74.conf (created at first run). For better results, I modified these settings:

```
[cpu]
core=dynamic
cputype=auto
cycles=max
cycleup=100
cycledown=100
...
[joystick]
joysticktype=fcs
```
- Games are expected under the directory roms/pc, a directory by game (with .pc extension) - this allows to give the name you want to the game and improve the scrapping !
- A file dosbox.bat is mandatory in game directory and must call the command(s) needed to launch the game
- An optional dosbox.cfg can modify specific configuration items (mapper is supported too)
- If you want a key mapper specific to your game, first add 

```
[sdl]
mapperfile=dosbox.map
```

to you configuration file and then call the mapper in game. If not, you will change the global mapper file.

Example of game structure for "Rick Dangerous" game:

```
roms/pc/Rick Dangerous.pc/
roms/pc/Rick Dangerous.pc/dosbox.cfg
roms/pc/Rick Dangerous.pc/RICK.COM
roms/pc/Rick Dangerous.pc/RICK.TND
roms/pc/Rick Dangerous.pc/SIMCGA.COM
roms/pc/Rick Dangerous.pc/RICK.CGA
roms/pc/Rick Dangerous.pc/RICK.EGA
roms/pc/Rick Dangerous.pc/dosbox.bat
roms/pc/Rick Dangerous.pc/dosbox.map
```

Still to improve:
- The mapper (which is opened by pressing CTRL-F1) has ugly colors (dark blue on black screen) ... must be a palette problem but haven't still found the workaround.

nadenislamarre:
hum, why not automatically suggest file with extension : .com, .exe and
.bat ?
Personnally, i would remove all .exe which are not games.

this is a typycal dosbox directory (the one of my childhood):
http://nadenislamarre.free.fr/c.zip

2016-03-04 9:02 GMT+01:00 Laurent Merckx notifications@github.com:

lmerckx:
Hello Nicolas,

I realize that it is not what is done generally in emulationstation but I found three advantages with this method:
1) the directory can have any name - which is more comprehensible for user and for the scrapping (not sure that the scrapper will always find the game with the .EXE file)
2) it is easier to select a game: you only have to select the directory (and not the directory, then the game file)
3) some games need to launch multiple executables (to execute driver, to mount CDROM directory, ...)

For the 1st and 2nd reasons, I also patched my ScummVM instance the same way: directory with .scummvm extension (with that, I can scrap all my ScummVM games without problem)

substring:
@lmerckx 
the problem this way is that users need to create an autoexec.bat each time they copy a file. That's something that we can discuss later, dosbox won't be added to rb-4.0.0 as we're sorting out the beta bugs for now
But I agree with you, it's nicer when you don't need to launch something more inside a folder. I was sure you made the same with scummvm ;)

nadenislamarre:
note that create a .bat for each it like making a white list (in a complicated way however).
since 4.0, es supports blacklisting,
so it would allow a very easy way, without removing the possibility to create a .bat if really needed.

lmerckx:
As requested, I split this dosbox PR to create a specific PR for SDL1 patch and another for configgen changes.
I also add static files: dosbox configuration, es_systems.cfg and addon for simplelight theme.

digitalLumberjack:
That is what i call a complete PR.
Thanks you !
Let's merge !

PR #244, opened 2016-03-14
Make sure custom.sh is executable

matthuisman:
If the user adds a custom.sh via SAMBA, it won't get the executable permission.
This fixes that and other cases where the user may forget to make it executable.

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S99custom
+++ b/board/recalbox/fsoverlay/etc/init.d/S99custom
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-test -e "/recalbox/share/system/custom.sh" && /recalbox/share/system/custom.sh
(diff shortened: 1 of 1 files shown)


nadenislamarre:
i dont validate.
samba is by default open :-(
people writting sh scripts should test and be able to  make it executable.

substring:
I agree with @nadenislamarre
What if they download a script called "superultimaterecalbox.sh", put it on the smb share and didn't read the script that contains just " rm -rf / 2>/dev/null"
That's a big security hole

nadenislamarre:
note that open samba is almost the same issue :
- copy on samba in system/.es/es_system.cfg and replace the python by rm
  -rf /
  ...
  ;-(

matthuisman:
Yes.
I didn't think of the security implications.

But, to be honest - anyone with access to the SAMBA share could cause a lot of damage.
Just deleting the share folder would by itself be enough probably to cause non-boot.

Maybe an idea to look into allowing password to be set for SAMBA access?

PR #259, opened 2016-03-20
omxplayer integration

LaurentMarchelli:
- [X] You added the changes in CHANGELOG.md
- [X] You choose the right repository branch to make the PR
- [X] You described the PR as below

Changes :
- Added omxplayer to be able to launch video from the command line.
- The version si based on the popcornmix version supporting fmpeg 2.8.1 not the last one (3.0). 
  Recalbox fmpeg version is the 2.8.4.
- Changed S02splash to run existing video instead than the splash screen.
- Add a quick video sample ... sorry I'm not a video designer, it was my first Blender experience.
- If we want to use an introduction video ... it has to be short ... very short mine is 20s long, and most of the time I can see only the fourth starting seconds before EmulationStation starts.

Creating a contest for video designer to select the best video that match the recalbox theme would be a nice idea. Wouldn't be ?


nadenislamarre:
You mean : a video at each boot ? It's very short (and could be faster again)
But in S10share ; upgrade part, it could be nice (but the % would be removed)
It is possible to used slideware for the moment.but nobody suggested something for the moment. Things like demo/captures of som games, tools recalbox is based on to thanks people, or new features or existing features

digitalLumberjack:
I like the idea :)

LaurentMarchelli:
Yes, I thought initially to an introduction video as it was done by Steve Smith on RasPlex & RetroPie Dual Boot. (see https://www.youtube.com/watch?v=jFP6THVNLWY)
But, as Recalbox definitevelly boots faster than Retropie does, we have to adapt.
For the update process, omxplayer support both layers and subtitles, so I guess it will be possible to dynamically draw version number and progression status over the video.

digitalLumberjack:
@ian57

ian57:
Thanks Digital, will put an eye on that.

A short video you said ;) https://youtu.be/5WmUboGAqxQ

LaurentMarchelli:
@ian57 
Definitively amazing ! u rock !

ian57:
@digitalLumberjack : omxplayer is already added in 4.1. I did nothing. Who did that... the mk file seems chinese to me ;)

@LaurentMarchelli : an existing blender intro modified to fit with recal ;)

digitalLumberjack:
@ian57 check the commit summary, it's @LaurentMarchelli that added omxplayer in this PR :)

ian57:
Saw that in the last seconds...

ian57:
@LaurentMarchelli you wrote the mk file from scratch... or did you adapt a existing one?

LaurentMarchelli:
I worked from several materials, I took the original package from https://github.com/Metrological/buildroot/tree/master/package/omxplayer.
But it didn't worked, mainly because of libraries api name changed, so I tried to fix the problems going on https://github.com/huceke/omxplayer.git and I saw that the version is now maintained by https://github.com/popcornmix/omxplayer.
The package can now be updated, for the moment, I fixed the popcornmix version to 1cd53be67b6be027cb1c70d20582b851d18223a9 because it is the ffmeg version 2.8.1 and we use the version 2.8.4.
I tried to build with popcornmix  HEAD, it worked. But, as the HEAD is in ffmpeg 3.0 I didn't want to take the risk.

LaurentMarchelli:
@ian57 
If u have any trouble, let me know, I may have look to fix it.

ian57:
thanks for all these informations... will try to make it work. Still compiling the last 4.1...
perhaps we could update to ffmpeg 3.0 in recal?

digitalLumberjack:
ffmpeg 2.x is required for many packages, be warned :)

LaurentMarchelli:
I'm not so sure it a good idea ffmeg is selected by : BR2_PACKAGE_KODI and severals others packages, I was checking this. Let's Kodi doing the update. ;-)

ian57:
Ok... so Laurent, you did the right choice ;)

LaurentMarchelli:
Trying to not break everything is often a good choice. :-D
Is your video available on git, I didn't extracted youtube video yet ?

ian57:
Here it is... https://github.com/ian57/recalbox-buildroot/tree/rb-4.1.X/board/recalbox/fsoverlay/recalbox/share_init/video

but I think we can do better :)

LaurentMarchelli:
Is it possible to have a transparent background ?
If yes, try this :
omxplayer -b --layer 10000 "/recalbox/system/resources/splash/recalboxIntro.mp4"
It will be played over EmulationStation.

LaurentMarchelli:
I guess on the final version the format will change to 720p or 1080p.

LaurentMarchelli:
Already very nice, u can run omxplayer from a terminal, the video will be displayed on the hdmi.
Let me know, if I can help.

nadenislamarre:
/fsoverlay/recalbox/share_init/video :
share_init is more for user data initialisation such as default recalbox.xml or kodi configuration, or default directories in the user directory

The choice : /recalbox/system/resources/splash/
is better.

we lack a directories naming, so, almost anything like this is valid.

ian57:
Ok thanks.... did not know where to put that ... so... 
I will do the modifications.

ian57:
Location changed to /recalbox/system/resources/splash/, generating the new SD image for tests.
I read that few video file have alpha channels (mov)... Why a transparent background is needed?

LaurentMarchelli:
We have the choice to make the video in front of the EmulationStation display (--layer 10000) or behind it:
- If we choose behind, the video is a bit too long, the Emulation splash screen will cover it before the end.
- if we choose to put in front of, with a transparent background your box will disappear in the EmulationStation menu.
  It could be an idea.
  I have a rpi3 boot and root image version 4.1 ready if it helps.

ian57:
Ok I understand better. I actually compute a new images sequence with alpha channel in 1920x1080. After I will try to create the transparent video in mov format to get the transparent background. Stay tuned. My pi2 image with video included  is ready... but no time to test!

nadenislamarre:
- If we choose behind, the video is a bit too long, the Emulation splash
  screen will cover it before the end.

why not include this in a script called : splash.sh
and add killall -9 splash.sh at the S31es script.

ian57:
Blender cannot export video with transparent background, openshot neither. As far as I know, to do that, I need to use AfterEffect or some other Adobe software that I don't own. I use mostly free software.

@nadenislamarre if you kill omxplayer during the video.. the results is the same... video will not finish and ES will be displayed.
Perhaps ES should wait the end of the video before starting?

nadenislamarre:
are we speaking about increasing the boot time ?

ian57:
In this case... no of course... so video is not a good idea ...

nadenislamarre:
isn't it nice even if we cut it when es start ?
an other idea is to remove the es splash screen to get more time for a
video.
think to upgrade process too, it takes 5 minutes to upgrade letting time to
watch videos ;-)

ian57:
I was thinking of a video with a looping part...  one full loop plays before es starts, and if es is delayed, it loops until it starts. I'm working on a thing like this.

If we can estimate the boot time on each  plateform, we can adapt the length of the video.

substring:
It depends on sd card speed, pi processor, overclocking, might be too complicated.

My very first idea was like @nadenislamare : start the splash with a SXX and kill it when S31 is launched

LaurentMarchelli:
The other solution is to use transparency, the video is played on the top layer and EmulationStation appears loading in background during the video.
If we increase the alpha blending from no transparency to full transparency during the video and remove the EmulationStation splash screen, then the recalbox in the video will vanish in the Emulation station menu.
In this case, we have no more issue of timing.

LaurentMarchelli:
oops sorry, I missed a message. :-(

LaurentMarchelli:
@ian57 Try this :
omxplayer --layer 10000 --alpha 128 "/recalbox/system/resources/splash/recalboxIntro.mp4"
When recalbox is launched.

LaurentMarchelli:
The best value to solve transparency problem seems to be :
`omxplayer --layer 10000 --alpha 224 "/recalbox/system/resources/splash/recalboxIntro.mp4"`

ian57:
I saw the recalboxmov video in /recalbox/system/resources/splash/. Is if a transparent background video ?

Arglll BR2_PACKAGE_OMXPLAYER is not set in recalbox-rpix_defconfig!!! -> not compiled and not in my SD grrrr

ian57:
Ok, i tried omxplayer --layer 10000 --alpha 224 on my rpi3 4.1 recalbox :) ... that's cool.. I think we have something here ;). Good job

LaurentMarchelli:
No, I don't think so.
Omxplayer natively support alpha-blending with the parameter --alpha you can set up the transparency gradient.

Envoyé de mon iPhone

LaurentMarchelli:
Another alternative could also be, create the video with EmulationStation splash screen as a vidéo background. 
As the splash screen is in the vidéo, you have the complete control on transparency during the animation. And at the end, your animation just disappears in the splash screen recalbox.
And we do not need to remove the Emulation splash screen.

LaurentMarchelli:
In this alternative, it will not be necessary to use transparency as we cheating with the splash screen.
We could also detect, joystick or keyboard input to stop the Vidéo before the end, as soon as the EmulationStation is loaded.

ian57:
Here is my omxplayer line : 

`omxplayer --layer 10000 --alpha 224 --aspect-mode fill --subtitles "/recalbox/system/resources/splash/subtitle.srt" --align left --no-ghost-box "/recalbox/system/resources/splash/recalboxIntro.mp4"`

with the file subtitle.srt  containing the update.done content

```
1
00:00:00,000 --> 00:00:07,000
v4.0.0 2016/03/23 10:10
```

and that's cool

nadenislamarre:
"We could also detect, joystick or keyboard input to stop the Vidéo before
the end, as soon as the EmulationStation is loaded."

are we speaking about the bootsplash feature ?

LaurentMarchelli:
It seems to be, I'm impatient to try it tonight.

Envoyé de mon iPhone

LaurentMarchelli:
Yes, we are trying to take care about your concerns over splash screen and boot duration, to be able to make a proposal that fulfills yours needs.

Envoyé de mon iPhone

ian57:
here is the line in the S02splash file

`omxplayer --layer 10000 --alpha 224 --aspect-mode fill --subtitles "/recalbox/system/resources/splash/subtitle.srt" --align left --no-ghost-box "/recalbox/system/resources/splash/recalboxIntro.mp4"`

here is the results https://youtu.be/1GXOV-Q2QE0

need to generate automatically the srt file from update.done

ian57:
Ok done on the fly (thx subs for the idea) in /tmp file : 

`more CreateSubT.sh 
#!/bin/bash
cat > /tmp/subtitle.srt << EOF
1
00:00:00,000 --> 00:00:02,000
EOF
cat >> /tmp/subtitle.srt < /recalbox/share/system/update.done
`

and in the S20splash

`# more /etc/init.d/S02splash 
#!/bin/bash
do_start () 
{
    if test -e "/recalbox/system/resources/splash/recalbox.mov"; then
        /recalbox/system/resources/splash/CreateSubT.sh
        omxplayer --layer 10000 --alpha 224 --aspect-mode fill --subtitles "/tmp/subtitle.srt" --align left --no-ghost-box "/recalbox/system/resources/splash/recalboxIntro.mp4"  
    else
        fbv -f -i /recalbox/system/resources/splash/logo-version.png
    fi
}
`

LaurentMarchelli:
Very, good work, I like it, it's amazing. I guess, you have been inside the Omxplayer help. ;-)
For srt update, two way of doing it :
- Compilation time. (I will have a check tonight)
- Update time : (version is stored inside /recalbox/system/recabox.version), if I remember well. A sed command can do the job.

ian57:
Here are my modifications : https://github.com/ian57/recalbox-buildroot/commit/92fad7221366835fb2f41cbf4cb6990970f7a195

ian57:
Yes, /recalbox/recalbox.version is better than /recalbox/share/system/update.done. will do the changes
Change done and another intro : https://youtu.be/KX6L-Ex0HKw

LaurentMarchelli:
I will update my clone end of the afternoon, I let you know. ;-)

nadenislamarre:
very nice !
yes, /recalbox/recalbox.version

LaurentMarchelli:
Hi guys,

First of all, I would like to thank you @ian57 for the amazing work he did, when you see my initial bullshit and the time it took, you just would want to say : Please don't play music if you're not a musician. :-) 

Then, I would like to present my point of view, it is just a proposal, as the decision is not mine and it is definitively yours. I would like to explain, my feeling on benefits and drawbacks I found for each video.

First video version V1, the explosion.
- The recalbox exploding from the stone, is definitively an amazing idea.
- It is breathtaking and energizing, as the recalbox is.
- It's just telling you, now we are here, just hang on if you can !
- Sorry I definitively love it. :-)
- But, I agree, this video cannot be cut before the end, and reducing the boot delay is definitively a high priority.
- Then, when we will reduce the delay, even with the transparency it could by annoying ... Even if, I definitively love it. :-D

Second video version V2, the neon.
- The idea is cool, it is just the openning of the amusement acarde.
- After the light is on, because it is looping, it can be cut at any time, and it definitely complies with reducing boot delay.
- However, I would like sound effects, starting with specifics neon clicks, then maybe a game machine starting sound, something looping as the video does.
- However, because it is looping, no way to know when it will finish. So, it should be only used for short delay ... but thank's to you, booting is already a short delay. :-)

What could be the final approach :
- Start with the version v2.
- Reuse the version v1 for upgrade, it could make sense, as the new version explode from the stone to enter in the system.

What we could improve :
- First of all, if we remove the splash screen, EmulationStation must be a part of the video, I guess you will agree, they definitively deserve it. For instance, it could be another background flashing neon.
- We can change the version number's font, to have something more fun.
- May be, have the version displayed longer.
- Create the subtitles during the build to avoid this process, each boot time.
- Use the GPU instead than the CPU, with the hardware acceleration (-w), but the tests I have made was not good yet ... too fast. :-)
- For the boot, I would like to start the video even sooner, I will have the look this week-end to see if there is any way.

Problems we have to fix :
- If we play a video during upgrade, it must be copied in memory before upgrade. (/tmp).  
- Version number does not appears on my RPI3 with full HD monitor, but I din't rebuild recently, just copied the files from @ian53 repository to my rpi.

That's all folks :-D

LaurentMarchelli:
@ian57 
Take care, you have a bug in /etc/init.d/S02splash :
- You check against recalbox.mov
- You run recalboxIntro.mp4
  I guess we will have to use a variable to avoid this.

substring:
Why not add the spkash video as a recalbox.conf parameter and stire videos on /recalbox/share ? Tvis way people can customize it without too much effort

ian57:
@LaurentMarchelli Ok for the bug... but as @substring said, it shoulb be better to lett the user put is own video, and make the right tests in the S02splash scripts. It is the next work. @digitalLumberjack told me the same.

 Thanks the feed back on the video @LaurentMarchelli . The V1 is partially my work. I modified a free blender intro from https://www.youtube.com/user/PartnersInCrimeGFX/videos. As you said : "Please don't play music if you're not a musician" :D. And I'm nor really a graphist... but I like and use Blender for severals years now... so I got some skills ;).

The V2 is mine and, the soundtrack was made in 30minutes.. so it is really poor... need to be enhanced.. I think i will make a V3 with only the Recalbox title and the rotating button... 

For video duration, I can "accelerate" the framerate to reduce length.

LaurentMarchelli:
Let the user choose the splash video can also be done without inserting a parameter in recalbox.conf.
In S02share, we just have to check if there is a video in share/system/splash before checking in /share_init/system/splash and during the upgrade create a link in /share/system/splash/ to the splash video.

LaurentMarchelli:
@ian57
I think the maximum framerate must be written in the video to allow a good hardware support.

nadenislamarre:
> > For the boot, I would like to start the video even sooner, I will have
> > the look this week-end to see if there is any way.
> > It will be hard to be sooner :
> > the bootloader start linux which start /sbin/init which start the
> > /etc/init.d scripts. The best place would be init, but keep it simple ;-)
> > 
> > Create the subtitles during the build to avoid this process, each boot
> > time.
> > yep. It's done here :
> > https://github.com/recalbox/recalbox-buildroot/blob/rb-4.0.X/board/recalbox/recalbox-patch-target.sh#L36
> > 
> > Why not add the spkash video as a recalbox.conf parameter and stire
> > videos on /recalbox/share ? Tvis way people can customize it without too
> > much effort
> > It's not as easy. And i would not recommend it.
> > 1) too much customisations makes the system more complicated to keep simple.
> > 2) the boot splash is the only place where recalbox can show its identity.
> > 3) you cannot access /recalbox/share before S11
> > 4) reading recalbox.conf during the boot takes 0.3s (because of the python
> > process loading)
> > 
> > More or less time
> > Please note that the boot process can take more or less time, so just think
> > how the video will happend in this situation :
> > - some external hard drive, takes up to 10s to because available to the
> >   linux stack once powered. (add the sleep 10 in S11 to simulate it)
> > - the 2 current boot slower are :
> > - we wait S01udev to finish (it takes around 3s), i'm not sure we have to
> >   wait, to in the future, this time could be reduced
> > - S26recalboxsystem is between 3 and 4s, and this is the current bigger
> >   problem. It's related to the fact that it reads all its values in
> >   recalbox.conf in python.
> >   an evolutio would be to generated a recalbox.conf.gen (a shell script to
> >   source) from the recalbox.conf each time recalbox.conf is modify to clearly
> >   improve the boot time.
> >   Which is not so easy while the rpi has no clock (you can simplily compare
> >   times, and things like checksum at the boot time is probably too long too).
> >   For the moment, i just parallelize in a maximum to get this time.
> >   ES takes a long time (es bootsplash) because (i've not analysed, just a
> >   supposition) of the xml when yo'uve a lot of games. If one day, we switch
> >   to sqlite, it will be almost immediate.
> >   But think about how things would happend if you remove 6s of the boot time
> >   (we could remove the video in this case however)

LaurentMarchelli:
Ok for starting the video sooner, ... it was it was challenge can forget for now ;-)

Now, let's try to have some proposal for more or less time :
- Less time : As the version 2 loops, we could kill the process as soon as EmulationStation is ready.
- More time : As the version 2 loops, we could loop until Emulation is ready.

Are these proposals acceptable ?

LaurentMarchelli:
@ian57 
Even if, you don't think you a musician, I definitively prefer you music than mine. ;-D
Sorry for the version 2, but I didn't had a sound or music at all. :-(

So guys, what do we do ? Do we should continue investing time on this subject ?
Now, it does not seems really clear for me, would you please take a stand on this point.

nadenislamarre:
Even without that, it is possible acceptable.
Take 2 or 3 seconds more is not a problem even if of course i would prefer
to remove that
You can do too: (sleep 2 ; killall -9 omxvideo) to kill in the worst case 2
seconds after s31es start.
I just mean : just take care about these two possibilies while the boot
time is not the same for everybody.
Le 24 mars 2016 19:18, "Laurent Marchelli" notifications@github.com a
écrit :

LaurentMarchelli:
@nadenislamarre 
Ok, we gonna try our best to fullfill requirements.

@ian57 
Hi Yann, I will contact you by email to try to synchronize on last issues during the week-end, if you agree.

substring:
Why don't you join us on irc ?

nadenislamarre:
2 other things
- first boot or when you switch on a usb key make the boot longer because
  of the copy of share
- fsck can happen on fat32 usb key if badly unplugged ; this can take time
  so it would be nice that after 15sec for example that the video stop and
  that the black screen with the logo is displayed because it displzys yhe
  fsck percentage
  Le 25 mars 2016 09:43, "Laurent Marchelli" notifications@github.com a
  écrit :

substring:
Can't there be a loopable part if the video ? Like a logo glowing, or buttons spinning like in the 2nd video ?

Just thout abt sthg : can't apply the video during upgrade process as the % is dynamically estimated

LaurentMarchelli:
If we use the version 2 it was also my idea, split the video in two parts.
- One without loop. The neon effect.
- The second part, the caroussel, that can loop until EmulationStation is started.
- We have to take care of video and sound during the change and see if there is no side effects.

ian57:
test Videos are ready... mailed to you this evening ;). There is sound in the 2 videos.

nadenislamarre:
Hi

Is there a way to disable the sound ? it's the 2nd time i get it when
starting my tv, and i'm looking for a way to disable the sound while it
makes my wife crasy.

ian57:
try to use the -n option with -1 to disable audio : Setting audiotrack to -1 disables the audio decode. 

`omxplayer -n -1`

LaurentMarchelli:
@ian57 
Ok I got it, sound works perfectly now on my rpi3.
Good news, looping with --loop also works on the second video.
More details by mail.

@nadenislamarre
Did you tried ... the remote mute button ? <:oD
Otherwise, we are fully open to a better sound ;-)

nadenislamarre:
the /boot/recalbox-config.txt could include an option for muting ;-) it is
available from the boot start.
it's not a problem of which sound, but depending on the last way of earing
the tv, the sound is loud, very loud, ...and it's suprising each time. too
short to find the remote.
anyway, i could add a such option ;-)

an other thing. i've kodi by default on start, for the moment, the video
loops strangely, i dont known if thats linked. i'll check why later.
Le 28 mars 2016 22:53, "Laurent Marchelli" notifications@github.com a
écrit :

LaurentMarchelli:
@nadenislamarre 
For sound volume, could you please try to add --no-boost-on-downmix parameter to omxplayer command line.
For Kodi, I guess it is a different problem, the video is stopped before 15s because emulationstation creates the /tmp/emulationstation.ready flag.
In case of starting with kodi, I suppose your are running the video until the timeout expires. (15s)
At the moment, we are still tuning the video mixing between first and second video, strange transition is not really surprising. 
However, if you see the transition only with kodi, it's because video is not stopped when kodi is ready and this is a valuable feedback.
Would you please give me some input on how the selection is done at the boot, so I will fix it.

PR #277, opened 2016-03-28
Fix scummvm patch files.

pasnox:


digitalLumberjack:
@rockaddicted ok ?

rockaddicted:
:+1:

PR #291, opened 2016-03-30
New video version and splash video now stopped when kodi is started

LaurentMarchelli:
- [X] You added the changes in CHANGELOG.md
- [X] You choose the right repository branch to make the PR
- [X] You described the PR as below

Changes :
- New video, with new musics, thank's to @ian57
- Move back to only one video, the simpler the better.
- Video is now stopped when kodi starts instead than emulationstation.
- Video length is 30s in case we need to increase timeout (first boot or update for instance).


nadenislamarre:
Hi,

i don't find where do you create the /tmp/emulationstation.ready file.
I thought you would have created here in es :
https://github.com/recalbox/recalbox-emulationstation/blob/rb-4.1.X/es-app/src/main.cpp#L320

so that it is launch when es is ready, and just before it launches kodi.

substring:
@nadenislamarre I think he means when kodi is automatically started at boot instead of ES

LaurentMarchelli:
@nadenislamarre 
You're close ... main.cpp L367 when all kind of initialization si done and when emulationstation is waiting for user input.

@substring 
You're right, "/var/run/kodi.msg" is made by kodilaunmher.sh L43.

LaurentMarchelli:
@ian57 
Great thank's to the artist for this nice artwork and also for his patience.
It is a pleasure to work with u. Dream team ;-)

nadenislamarre:
by creating the /tmp/emulationstation.ready a bit ealier,
you could remove the dependance of /var/run/kodi.msg
/var/run/kodi.msg is a pipe for a watchdog i've created for kodi but it may disappear in the feature.

rockaddicted:
Great job guys, very impatient to test this feature.
Just for information, in 4.0.0 stable release, we'll remove the currents background musics because there is a copyright issue :-/
The recalbox community has provided us new open source musics to be included in recalbox by default.
So could it be possible to use one of them (or an other opensource music) in the background of the movie instead of mario's theme,  please ?

https://github.com/recalbox/recalbox-themes/tree/master/recalbox/recalbox/music

LaurentMarchelli:
@nadenislamarre 
Sorry, I think it is made is the right place for emulationstation, if we stop the video sooner we may see the emulationstation splash screen.
When you will review kodilauncher.sh, it is not a big deal to add "touch /var/run/kodi.msg", "touch /tmp/kody.ready'" or touch "/tmp/emulationstation.ready" as you like.
So, you would be able to adjust precisely when you want to stop the video.
Let me know when you will do it, it will be my pleasure to modify S02splash accordingly. ;-)

@rockaddicted 
Got it, I forward the message to the artist and I gonna update the video, please keep this PR open until we have the video with community sound.

nadenislamarre:
ok. I didn't now it was very precise.
Let's keep like it is for the moment then.
Thanks a lot for your job.

LaurentMarchelli:
Yes it is, there is still a strange behavior with kodi. 
But it is the kodi launcher behavior as it also appears with splash picture, I guess it is because video output mode is changed twice when running kodilauncher.sh.
So the kodi splash screen it not visible longer, it was a part of the strange behavior you detected.

nadenislamarre:
We can disable the kodi splash if you think it's better
via http://kodi.wiki/view/Advancedsettings.xml#splash (not tested)

~/.kodi/advancedsettings.xml
and then in share_init/system/.kodi

Nicolas

ian57:
Problem with sound if you use the Jack output for sound and a HDMI-VGA converter for display. No sound during the video. The HDMI output is selected by default. 

We need to force the audio device for omxplayer thanks to the -o option : 

`omxplayer -o local ... or omxplayer -o hdmi ...`

But we need to know which output is used!

LaurentMarchelli:
@ian57 
You're right, I gonna check.

digitalLumberjack:
For your concern, recalbox.conf `audio.device` is loaded here 
https://github.com/recalbox/recalbox-buildroot/blob/rb-4.0.X/board/recalbox/fsoverlay/etc/init.d/S26recalboxsystem#L32 to change the used device here : 
https://github.com/recalbox/recalbox-buildroot/blob/rb-4.0.X/board/recalbox/fsoverlay/recalbox/scripts/recalbox-config.sh#L230 with `amixer cset numid=3 X` where X is 0, 1 or 2.

ian57:
I changed in ES the used device... but, the video is loaded before with S02Splash... the configuration arrives to late. We need to read the configuration before. /var/lib/alsa/asound.state is not present on recalbox. Is there another file with the audio configuration?

video https://youtu.be/qs_VZIHGzv4

LaurentMarchelli:
We will do as we do for switch, directly sed the recalbox.conf not need to use recalboxSettings.pyc.
https://github.com/recalbox/recalbox-buildroot/blob/rb-4.0.X/board/recalbox/fsoverlay/recalbox/scripts/powerswitch.sh#L174

nadenislamarre:
but in S02, you have no access to share, then to recalbox.conf.

ian57:
So we need to write somewhere else the "audio" config to be able to read it at S02 ;)... but everything is in read only!!!! before share is accessible... An idea?

In debian, you have access to /var/lib/alsa/asound.state which contain the last config. But i found nothing similar in recalbox

nadenislamarre:
you can use /boot/recalbox-config.txt

mount -o remount,rw /boot
edit /boot/recalbox-config.txt
mount -o remount,ro /boot

pratically :

my guess is to edit /recalbox/scripts/recalbox-config.sh
in the same place, we set the audio and we call
preBootConfig/postBootConfig as you need.

LaurentMarchelli:
@nadenislamarre 
You're right, it's my mistake.
Here is the right answer : hdmi_ignore_edid_audio=1 in config.txt

LaurentMarchelli:
The complete solution step by step :

```
mount -o remount,rw /boot
echo "hdmi_ignore_edid_audio=1">>/boot/config.txt
reboot

```

Nothing else to modify.

LaurentMarchelli:
@nadenislamarre 
It is not as simple as removing the splash screen, I think it's because you forcing the video mode in configen and I guess there was a reason to do it. Right ?

substring:
@LaurentMarchelli you can set a custom video mode to any emulator in the recalbox.conf. Kodi is handled as an emulator

LaurentMarchelli:
@substring 
Thank's for the information, I saw it, but I just do not have the time to get into Kodi launching problem for now.
I just took care that the video didn't inserted any regression even with Kodi and that's ok.

digitalLumberjack:
Let's try that :)

PR #315, opened 2016-04-16
Rb 4.1.x nadenislamarre sshpassword

nadenislamarre:

Changes :
- Randomize the ssh root password
- The ssh password can be updated
- The ssh password is visible on the es interface (see an other pr)
- not securing missing part : samba, manager 

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S22rootpasswd
+++ b/board/recalbox/fsoverlay/etc/init.d/S22rootpasswd
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# /etc/shadow is dynamically generated from the password found in /boot/recalbox-boot.conf
+# the password is visible only in the es interface
+# or to people having already a ssh password via the command : /recalbox/scripts/recalbox-config.sh setRootPassword xyz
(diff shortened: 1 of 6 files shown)


nadenislamarre:
Some points :
- securing ssh without samba is not secure (you can modify allowed public keys)
- samba and manager will require the same password
- the password is saved in the /boot because i don't think it's nice to move it when sharing/copying usb keys (however, the password encoding is related to the pi hardware and a new one will be generated if not readable)
- the root password can be rewritten in clear in the /boot/recalbox-boot.conf file in case of emergency (so, from a computer)
- an option in system options => security will allow to enable/disable security, by default, off
  => security disabled should have the same effect as in 4.0, including setting the root password to recalboxroot
- the security parameter will not touch to kodi options. apple airplay control is enabled by default in kodi, without password. We should perhaps enable the http control (for remote phone applications) while some people don't enable it and don't understand why it doesn't work out of the box.

any other suggestion is welcome.

digitalLumberjack:
Do you want me to add the option in ES system options ?

nadenislamarre:
no.
i'll add things.
it's just that i'm doing it pr by pr to make less things to validate at a
time to make it easier for you (and for me).

digitalLumberjack:
Cool

PR #319, opened 2016-04-18
security : samba requires a password if system.security.enabled=1

nadenislamarre:

Changes :
- add a recalbox option : system.security.enabled = 0 by default
- if 1, samba requires a password (root password displayed in es => system)
- if 1, virtual gamepads are disabled

Unfortunatly, it seems that i can't disable dropbear public keys when it is 0.
To do : secure the manager

Next pr : add the clamcav virus scanner.
Todo : disallow the recalbox to startup until the scan is done (takes less than 20 minutes the first time, and about 4 minutes the next times).
The virus database takes less than 200mb, so, it may not increase a lot the disk requirements.
It increases a bit the psx iso file copying by 2 minutes too, which is nothing compared to the security it provides.
Todo : put  system.security.enabled to 1 until the clamcav database update is finnished and the new virus definition have been used to rescan the drive.

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S35securepasswd
+++ b/board/recalbox/fsoverlay/etc/init.d/S35securepasswd
@@ -12,6 +12,11 @@ then
     MASTERPASSWD=$(/recalbox/scripts/recalbox-config.sh getRootPassword)
 fi
 
+# secure ssh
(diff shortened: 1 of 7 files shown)


digitalLumberjack:
nice

PR #331, opened 2016-04-27
DosBox - SDL2 (2)

lmerckx:
Sorry, some issues with my previous PR :
- Bad revision in dosbox.mk: 3970 instead of 3980 -> resolved
- Add default configuration file in fsoverlay


lmerckx:
There is also a PR in configgen repository to add the GL driver in environment variables.

PR #333, opened 2016-04-30
Rb 4.1.x nadenislamarre brmerge3

nadenislamarre:
- Upgrade to linux 4.4 + fix buildroot issues about rpi-firmware.
  - Fix gcc5 not set for rpi1 and rpi3

Limitations :
I've not tried to fix the 3 limitations, it seems quite simple problem to fix.

1) the current patches have not been copied from 4.1 to 4.4 because they fail
linux-4NES4SNES-hid.patch
linux-4NES4SNES-hid-quircks.patch
linux-xpad-xboxone-support.patch

2) recalbox-api doesn't compile. But i don't know what it is for, so, i skipped.

3) the sound is working correctly in kodi and at boot, but in es and in games, i get no sound. I don't know what it is related too.


substring:
@nadenislamarre tbh, we'd better consider upgrading to 4.4 later, in a different PR. This one should just update buildroot

nadenislamarre:
linux-4NES4SNES-hid.patch
linux-4NES4SNES-hid-quircks.patch
are already applied in 4.4. not need these patches anymore.

rest to do :

1) linux-xpad-xboxone-support.patch

2) recalbox-api doesn't compile. But i don't know what it is for, so, i skipped.

3) the sound is working correctly in kodi and at boot, but in es and in games, i get no sound. I don't know what it is related too.

nadenislamarre:
for the xbox patch,
the code changed a lot.
some part seems already applied.
some are not.

it seems that the patch is in linux 4.5, not in 4.4 however.
http://lxr.free-electrons.com/ident?v=4.5;i=xpad_presence_work

i dont know where we found the version for 4.2
we should search the same for 4.4
it seems to come from here : https://github.com/paroj/xpad
there are a lot of difference between the xpad.c from 4.2 and the one from 4.4 and the one from 4.5

i would suggest to take the xpad.c from the 4.5 while the functions structures are the same, just the content differs.
This drivers evolves a lot.
or to test the driver as it is in 4.4 to see if it works already correctly. i don't know from when is our patch, from when is the 4.4 patch.

Anyway : how as xbox pads to make some tests ?
i suggest to go in 4.4 for the recalbox 4.1 for tests before first beta, while the rollback is easy just the kernel version.

digital, subs, rockaddicted ? you still refuse my pr or do you allow to merge ?

substring:
I don't mind moving to 4.4. Anyway i'm afraid it's mandatory since dtbo were committed at the same time. The RPiF gave up the 4.1 branch so, sooner or later, we have to switch to it

Xpad hs something to do with xbox one pads i think. @rockaddicted should know about that.

That sound thing is rather annoying :/

Anyway, the final word goes to @digitallumberjack

nadenislamarre:
the sound will be fixed. i've to do a full rebuild to fix it.

substring:
Ok, i'll test your pr and do a clean rebuild

nadenislamarre:
link for xpad.c
it is the one i'll try to apply for the odroid kernel.

https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/plain/drivers/input/joystick/xpad.c?id=refs/tags/v4.5.5

digitalLumberjack:
Let's merge that :)

PR #358, opened 2016-05-28
add support for usb network interfaces

damiencourousse:
Everything is in the title.  I would like to get a network access via a USB network interface, so that I can share the network connection of my smartphone with the recalbox box.

This change allows usb network interfaces to be ifuped at startup.  

I am not that much familiar with buildroot, and maybe there is a better way to do this.  For example, it would be nice to be able to mount usb network interfaces after startup.  Kodi currently supports this, but I am not sure how to achieve this in recalbox.  I would happy to add more changes to support this: tips are welcome !!

thanks,
Damien

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Slide transition by default in ES
 - Power management switch support (power,reset and LED) for pin 3/5/6
 - Add ifconfig -a and /boot/recalbox-boot.conf in recalbox-support.sh
+- Add support for usb network interfaces
 
 ## [4.0.0-beta3] - 2016-04-19
 - Xarcade2jstick button remapped + better support of IPAC encoders
@@ -39,7 +40,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed system locales
 - Updated 8bitdo gamepads
 - Bumped to moonlight-embedded-2.1.4
-- Overclock set to none now delete lines in config.txt 
+- Overclock set to none now delete lines in config.txt
 - Improved keyboard encoders support
 - Fixed an issue concerning ISO loading taking too long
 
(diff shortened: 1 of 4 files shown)


digitalLumberjack:
Hi pal, thanks for your pr.
We are about to release 4.0, i cannot merge new feature in it. Maybe you could reopen a PR on 4.1 branch and close this one, and we could discuss about the code on the new pr ?

Thanks :)

damiencourousse:
hello,

Yes, I can open a new PR on 4.1.  But I would like to ask you for an advice before: in 4.1, `eth0` has disappeared from `/etc/network/interfaces`, and I see that the script `recalbox-config.sh` no longer takes the command `ethernet`.  How is the ethernet interface auto-mounted?  Would this be suitable for usb network interfaces as well?

thanks,
Damien

substring:
It's handled by connman on 4.1. Maybe @nadenislamarre can give more details

damiencourousse:
I will have a look, thanks.

damiencourousse:
Closing.  Thanks for your feedback!

It looks like my feature request will be supported by connman in the next release.  Currently, in version 4.0-beta3, networking via usb ethernet is already supported by the Linux kernel, but the network connection is not automatically mounted.

thanks,

nadenislamarre:
dont close.
as far as i remember i have enabled only some connmann feature. i have to
check the good one is enabled

Le dimanche 29 mai 2016, damiencourousse notifications@github.com a
écrit :

nadenislamarre:
connman modules enabled :
Ethernet support
WiFi support
command line client

disabled:
WISPr support
Bluetooth support
loopback support
neard support
ofono support

do you known which module is required for what you want ?

damiencourousse:
hello Nicolas,

The ethernet module should be enough for my needs.
I had a look around on the internet but could not find better information for this, sorry.  

best regards,
Damien

PR #368, opened 2016-06-02
odroid xu4 : sdl2

nadenislamarre:
upgrade sdl2 from 2.0.2 to 2.0.3 on the odroid (required for moonlight)
I created the patch by getting the delta between the sdl2 odroid branch and the official sdl2 and cleaning it.
and i applied this patch over 2.0.3 while it's not intrusive.
I mainly add a video driver and a switch in the configure script.
Note that this patch is applied for the rpi too, but the switch is not applied because the check fails.


nadenislamarre:
ok, in fact moonlight requires sdl 2.0.4 for the xu4 (to use the sdl2 video output)
so, wait before applying this one.

digitalLumberjack:
Nice.

nadenislamarre:
the buildroot team confirmed to me that if sdl2 is not in 2.0.4 it is just because nobody has send a patch to upgrade it.
First i test a bit (i'm on 2.0.4 on my xu4), and i'll push it to buildroot.

nadenislamarre:
still don't validate this one
i've to test that moonlight works.
I've no server to test.

nadenislamarre:
i see you merged. anyway.
i'll do an other one.
the buildroot team suggested me to do like this for the patch :

$ grep PATCH configs/calao_qil_a9260_defconfig 
BR2_GLOBAL_PATCH_DIR="board/calao/qil-a9260/patches/"

to apply it only on the xu4, and not for the pi.
in other words :
"<y_morin> : We try to avoid platfrom-specific patches, yes."
<y_morin>  If that patch really can't be made generic (e.g. a #define), then you'd have to carry it in the board/xu4/sdl/your-changes.patch  and use that directory as global patch dir.

rockaddicted:
Since this PR, sdl2 don't build anymore. Tried to build today a fresh 4.1 for rpi3 : http://pastebin.com/VJkqkEYq

rockaddicted:
after added include missing I have this new error : http://pastebin.com/DMakrLfd

PR #375, opened 2016-06-09
DosBox: add hardware support for joystick up to 8 axis

lmerckx:
Allow to support hardware joysticks with 8 axis. 
Caution: Emulated joystick remains unchanged !

Only done for mapping purpose (all axis, hats and buttons are now mappable).


digitalLumberjack:
Cool :)

PR #377, opened 2016-06-13
Rb 4.1.x nadenislamarre xu4 netboot

nadenislamarre:
this pr contains the prerequisite kodi one because of the new shared fsoverlays.

in other word, it's now possible to put share, or just bios/roms/saves on a network point like a nas or a fr**box.
-    # /boot/recalbox-boot.conf examples :
-    # sharedevice=NETWORK
-    #
-    # Basic commands : sharenetwork_<nfs|smb><[0-9]>=<SHARE|ROMS|SAVES|BIOS>@<host>:<remote directory>:<mount options>
-    #
-    # sharenetwork_nfs1=SHARE@192.168.0.1:/Documents/recalbox
-    # or
-    # sharenetwork_nfs1=ROMS@192.168.0.1:/Documents/recalbox/roms
-    # sharenetwork_nfs2=SAVES@192.168.0.1:/Documents/recalbox/saves
-    # or
-    # sharenetwork_smb1=SHARE@192.168.0.1:Documents/recalbox:guest
-    #+    # Advanced commands : sharenetwork_cmd<[0-9]>=<command to run>
-    # sharenetwork_cmd1=mount -o port=2049,nolock,proto=tcp 192.168.0.1:/Documents/recalbox /recalbox/share
-    # or
-    # sharenetwork_cmd1=mount -o port=2049,nolock,proto=tcp 192.168.0.1:/Documents/recalbox/roms /recalbox/share/roms
-    # sharenetwork_cmd2=mount -o port=2049,nolock,proto=tcp 192.168.0.1:/Documents/recalbox/saves /recalbox/share/saves
-    # or
-    # sharenetwork_cmd1=mount.cifs //192.168.0.1/recalbox /recalbox/share -o guest


substring:
hey @nadenislamarre 
Well done for the network mounting stuff ! But I need a few explanations
- Why add networks shares per /recalbox/share/directory ? It has never been the case so far, you make things getting complex for no reason. Keep it simple : just mount the /recalbox/share as a whole, just like for USB keys !
- I definitely do not agree with the concept of the S11network-preconfig. If a S11 has to run some S30 and S45 script, there is a misconception problem
- what is that S16modprobe ? The kernel is supposed to load everything itself, so why force it ?
- S40network + S45connman : the stop part is bad. They must be stopped, network or not. Unmount FS before
- your printf should go to the recalbox.log or the system log

There is an overall start/stop problem : network should be started before mounting the share, so quite early in the init process. And as you noticed network can't be enabled over wifi because of the recalbox.conf location. Probably the solution would be to move wifi connection details to the recalbox-boot.conf but offering NFS/CIFS over WiFi may not be that wise afterall ...

The process order should be dbus -> network -> connman -> mount -> ES. I know this order will slow down the boot process. But starting init scripts in the right order is better than hacks calling some scripts that were supposed to run much later.

Looks like there are redundant commits between this pr and #376

nadenislamarre:
1) Why add networks shares per /recalbox/share/directory :
For several reasons.
a) i've several boxes, i want to share only the roms and the saves.
b) the target is to mount saves via googledrive, not the rest (for performances reason)
c) it doesn't make the code more complex.
d) for testers, when you have several cards, tha'ts really easier to not have to copy things each time

2) S11network-preconfig
I don't too. But it is kept simple.
and systemv doesnt allow too much configuration about dependancies.

3) what is that S16modprobe 
this is not mine. It already exists.
I just added some if $1 != start do nothing because currently, if you watch the shutdown process, this is very messy, a lot of things are done that should occur only at boot time, not at shutdown time.

4) dbus -> network -> connman -> mount -> ES
as you prefer.
personnaly, i prefer a mini hack, than 10 seconds more (or even more on wifi) to boot.
my opinion is that booting from the network require a special order, but that's not possible to configure with systemv. the current boot sequence is good, but doesn't allow to mount network share. replacing it by a worse one in case you have no mount share is not a good choice i think : i won't add restrictions because of software configuration limitations. there are 2 corrects boot sequence depending on options.
The umount part is currently done by inittab, but could be made in a init.d/script, but how to get the list of umount to do ? not simple too.
I've search several solution, and my conclusion was to prefer this mini script which is really simple.
For the others : doest have a sense to do ifdown at shutdown ? not sure.

PR #391, opened 2016-06-22
Update config.txt

substring:

Fixes #XXXX

Changes :
- Enable kernel 4.4 audio

### DIFF
--- a/package/rpi-firmware/config.txt
+++ b/package/rpi-firmware/config.txt
@@ -10,6 +10,10 @@ disable_overscan=1
 # comment to get jack audio
 hdmi_drive=2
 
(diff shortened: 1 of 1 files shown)


nadenislamarre:
note that /etc/init.d/Sxxmodules was added manually (by digit i think)
and loads files from /etc/modules, so it works in our case.
The question is to minimize/remove it.

substring:
Loading the module doesn't make the sound working ! The device tree node has to be enabled for the spund to work. Hence this PR.

The sound was broken since the upgrade to kernel 4.4. it's been reported many times inside the team. This pr fixes it, finally

PR #397, opened 2016-06-25
Dosbox - emulated mouse

lmerckx:
Allow to emulate a mouse (3 buttons) with joystick or keyboard, via the mapper.
Best usage is to map X/Y directions with an analogical stick, because mouse speed is proportional to stick thrust. Also usable with pad and keyboard but with uniform speed. 

Surely improvable ! Don't hesitate to leave comments ...

Ideal to play mouse-only Dos games (IE: Sim City 2000)


digitalLumberjack:
Seems great :)

PR #399, opened 2016-06-29
pygame : introduce the pygame (sdl for python) library for the upgrading system

nadenislamarre:
It allows to make nicer small framebuffer interfaces in a few number of lines.
In the future, it can allow:
- a screen to choose the language the first time
- a screen for the rsync usb key if es needs to be stopped
- a small image viewer to replace the not nice libretro-imageviewer
  ...


substring:
Why don't you discuss such upgrades with the team ?

Why remove sdl_image libraries ?

nadenislamarre:
i think there is a misunderstood on the pr.
I've not removed sdl libs. It is the mecanism of buildroot that the config files contains only leaf packages, not the dependencies. So, technically, sdl_image is removed from the config, but it is still here.
I've just added the sdl python binding. The recalbox system uses a lot python, but it was not possible to display things on the framebuffer from python.
The aim of this pr is to change the upgrade screen display.
I've tested and read about pygame these last days, i wanted to make a test on recalbox, and this is a suggestion. Feel free to discuss in this pr ;-)

ian57:
I tried to use pysdl2 with FB, but no success. Seems to be made for xorg use.

PR #400, opened 2016-07-03
img card : create the rpi .img at the end of the build process as it …

nadenislamarre:
img card : create the rpi .img at the end of the build process as it is done for the xu4


substring:
Great ! the next step is to sqhashfs them ? :)

nadenislamarre:
no, there is no really link between squashfs and .img,
i think about squashfs, but i don't know if that is what we want.
It makes cleaner and faster updates.
I'll make tests, i'll see.

digitalLumberjack:
Great :)

PR #403, opened 2016-07-07
exfat : add exfat support

nadenislamarre:
it's not a recommended format for linux, but it works, and it allows to
mount usb keys containing video for kodi for example.


digitalLumberjack:
Alright :)

PR #408, opened 2016-07-13
SDL_sound: patch to use SDL2 instead of SDL

lmerckx:
This patch allows SDL_sound to be compiled with SDL2 instead of SDL1.
It allows DosBox (now compiled with SDL2 too) to decode compressed audio files (OGG) in IMGMOUNT.

Indeed, it is not possible to use in a same application SDL2 and SDL_sound, compiled with SDL1.

In previous versions of RecalBox, SDL_sound was not used. In 4.1, the only dependence is DosBox. So, no impact expected !

For more information, please consult: http://www.vogons.org/viewtopic.php?f=32&t=34770&sid=ed1915709f4262832c27d886b3160a06&start=52


digitalLumberjack:
Thanks, i'll merge :)

PR #416, opened 2016-07-20
lutro libretro core

Laddyacris:
Add lutro extension for games.
Need to bumped lutro core to make compatibility
8/8 games ares works


paradadf:
Don't you need to bump the lutro libretro core to allow .lutro extensions as well?

Laddyacris:
When I try lutro game 3/8 works. 
When I use a recent lutro core libretro 8/8 games worked.

Mon repro ne s'est pas mis à jour apparemment via github desktop.
Il faut que je trouve comment réussir.

cette PR n'est pas valide. Grrrr.

PR #418, opened 2016-07-24
Rb 4 1 0 bump lutro

Laddyacris:
Update lutro Core


ian57:
On my 4.1 the makefile.patch does not work on the bumped release.

PR #427, opened 2016-07-30
Added bit and sampling rate info for emulation station bgmusic files

robsdedude:
Added some more explanation on the background music readme.txt

I heard only something like white noise in the emulation station menu due to an unsupported bitrate.


digitalLumberjack:
thank you :)

PR #454, opened 2016-08-24
Update SNES button mapping to match NES (A, B)

### LINKED ISSUE #1: Support to Wireless dongle
From mgoulartBR

Hi again, I was thinking about a support for wireless dongles (like Broadcom and Atheros). The reason to ask it, it's because all the time that have any update, I need to open my arcade cabinet, remove the RPi, get some cables and plug them all on my TV. And after the update, unmount all the cables and remount RPi on my arcade again.

I think the challenge is integrate an interface to select a wifi connection on Emulation Station

munskey:

Fixes #1

Changes 
- SNES button mapping updated to match NES buttons (eg: NES A = SNES A, NES B = SNES B)
## \- Allows NES and SNES controllers to share the same controller port using a simple plug one-to-one adapter without requiring unique ES controller profiles

### DIFF
--- a/package/gamecon_gpio_rpi/gamecon_gpio_rpi.c
+++ b/package/gamecon_gpio_rpi/gamecon_gpio_rpi.c
@@ -422,7 +422,7 @@ static void gc_gcube_process_packet(struct gc *gc)
 #define GC_NES_LATCH	0x800
 
 static const unsigned char gc_nes_bytes[] = { 0, 1, 2, 3 };
-static const unsigned char gc_snes_bytes[] = { 8, 0, 2, 3, 9, 1, 10, 11 };
+static const unsigned char gc_snes_bytes[] = { 0, 1, 2, 3, 8, 9, 10, 11 };
 static const short gc_snes_btn[] = {
 	BTN_A, BTN_B, BTN_SELECT, BTN_START, BTN_X, BTN_Y, BTN_TL, BTN_TR
 };


substring:
mmmhhhhhhhhh ... And what if we upgrade gamecon to 1.2 ? does it solve this issue ? I was told it also solves the PS driver that can't handle more than 1 pad at once
see https://github.com/recalbox/recalbox-os/issues/1057

munskey:
Gamecon 1.2 doesn't solve the issue, as RetroPie uses 1.2 and the code still exhibits the symptom.

To clarify, my pull request isn't really an 'issue', but more of a feature request due to an easy oversight by the original gamecon developer.

The NES buttons are a subset of the SNES buttons both on a physical button layout and on an electrical level they are encoded the same and operate the same.  Programmatically the gamecon driver assigns arbitrarily different integer values in the NES and SNES buttons arrays.  Meaning, NES A is assigned the integer value of 0 and SNES A is assigned integer value of 8 for their respective arrays.  There may be a reason for this difference, but I was unable to determine it.

I want the controller types to use the same integer array numbers so that I can simplify the controller hookups by using only one type of physical connector on my case (NES or SNES), then, for example, when I want to connect a SNES controller to an NES port I can (using a homemade one-to-one physical NES-SNES adapter) and the button mappings won't be thrown off.

Hopefully that makes more sense ;)

substring:
Thank you for the loooong explanation :)

The only thing that concerns me is if it changes the button order from a linux point a view. For example, they would not trigger the same events as before (A and B swapped). Can you give me your regard on that ?

munskey:
You can swap or remap the buttons on a linux level, retroarch, etc.  I think that would require the user to manually notify the system that, "hey, I'm using controller type X now" for it to perform the remap after the gamecon driver output.

What I am proposing is transparent and intrinsically automatic to the end user since recalbox will be seeing the exact same output from the gamecon driver (without and intermediate modification) regardless if you are using a SNES or NES controller, i.e. no manual intervention is needed, nor should it be since one is a subset of another.

Perhaps a better route would be for me to send a request directly to Marqs to ease future updates.  I don't see a github for it, only his site https://www.niksula.hut.fi/~mhiienka/Rpi/

substring:
Alright. I'm not the one who decides, but this is what i can do. There is an ongoing PR for the 1.2 driver version : i'll add a patch with your submission. We're terribly lacking testers with gamecons, so digi is a little reluctant to it ;)

digitalLumberjack:
build board rpi3

digitalLumberjack:

digitalLumberjack:
ok to test

digitalLumberjack:
test this please

PR #478, opened 2016-09-13
dosbox: add readme files, and the pc directory

nadenislamarre:
i kept the name "pc" as in the emulationstation system config file, but perhaps that all should be renamed "dos".

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S12populateshare
+++ b/board/recalbox/fsoverlay/etc/init.d/S12populateshare
@@ -33,7 +33,7 @@ for FILE in music bios extractions kodi saves screenshots \
             system/.config/lirc/lircd.conf \
             system/.emulationstation/{es_input.cfg,es_settings.cfg} \
             system/{.hatari,.kodi} \
-            system/configs/{fba,moonlight,mupen64,ppsspp,reicast,retroarch,shadersets,xarcade2jstick} \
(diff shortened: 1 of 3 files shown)


board/recalbox/fsoverlay/etc/init.d/S12populateshare
```diff
@@ -33,7 +33,7 @@ for FILE in music bios extractions kodi saves screenshots \
             system/.config/lirc/lircd.conf \
             system/.emulationstation/{es_input.cfg,es_settings.cfg} \
             system/{.hatari,.kodi} \
-            system/configs/{fba,moonlight,mupen64,ppsspp,reicast,retroarch,shadersets,xarcade2jstick} \
+            system/configs/{fba,moonlight,mupen64,ppsspp,reicast,retroarch,shadersets,xarcade2jstick,dosbox} \
```
substring, line 36: Juste a note : we should really find another way for `~/configs` creation/upgrade. Would like feedback on https://github.com/recalbox/recalbox-os/issues/1043
And so far `/recalbox/share_init/system/configs/dosbox` doesn't exist, but a `.dosbox` in system. Should settle these down

substring (approved):

substring:
@nadenislamarre I agree with dos instead of PC

nadenislamarre:
i'll update from pc2dos once this pr is validated.

PR #479, opened 2016-09-15
x86: fix the keyboard

nadenislamarre:
pressing a key on a keyboard to display a letter is a complex action ;
mainly to support special chars, like keys that you can't have by pressing
a simple key, like chineese keys.
Under X, there is generally a server, an input method server, such as ibus
to handle that.
Unfortunally, such server have a lot of dependancies, and is not by default
in buildroot.
If no such server exists, kodi fall back to a more basic system not supporting
some keys. SDL requires a such server to run with X11 ; except if your X11
implementation doesn't support UTF-8... or if you make it think it doesn't via
a basic patch.

... several readings, but finally only one line to fix ;-)
note that special chars, like french accents doesn't work. I think an IM
is required to get them working.


digitalLumberjack:
thx

PR #480, opened 2016-09-16
1st recalbox.conf update try

substring:

Fixes #XXXX

Changes :
- compare `/recalbox/recalbox.version` with `/recalbox/share/system/logs/lastrecalbox.conf.update` and check if an update is necessary
- browse through `/recalbox/share_init/system/recalbox.conf` for non-commented values
- check if the property has to be forced or skipped (see `forced` and `skipped` arrays) or just missing in the user recalbox.conf
- if the property is missing, it is added
- log everything in the recalbox.log

The first check on the version is necessary as the whole process takes about 1.5s on a pi3. We don't want to increase the boot time when updates are not necessary.
There is an ugly `exit 0` if no update is necessary, that's why this section is at the very end of the script

### DIFF
--- a/board/recalbox/fsoverlay/etc/init.d/S12populateshare
+++ b/board/recalbox/fsoverlay/etc/init.d/S12populateshare
@@ -90,4 +90,50 @@ mkdir -p /recalbox/share/system/udev/rules.d
 rm -rf /run/udev/rules.d
 ln -s /recalbox/share/system/udev/rules.d/ /run/udev/
 
+# Update recalbox.conf
+cfgIn=$IN/system/recalbox.conf
+cfgIn=$IN/system/recalbox.conf
+cfgOut=$OUT/system/recalbox.conf
+rbxVersion=/recalbox/recalbox.version
+curVersion=$OUT/system/logs/lastrecalbox.conf.update
+LOG=$OUT/system/logs/recalbox.log
+skipped=(system. kodi. wifi. audio.)
+forced=(controllers.ps3.driver)
+
+# Shall we run the recalbox.conf update process ?
+currentVersion=$(cat $rbxVersion)
+lastConfUpdate=$(cat $curVersion 2>/dev/null)
+
+[ "$currentVersion" == "$lastConfUpdate" ] && echo "recalbox.conf already up-to-date" >> $LOG && exit 0
+
+echo "UPDATE : recalbox.conf" >> $LOG
+
+cp $cfgOut ${cfgOut}-pre-$(cat /recalbox/recalbox.updateversion)
+grep -v '^[#;]\|^$' $cfgIn | while read -r property
+do
+    skip=0
+    name=$(echo $property | cut -d '=' -f 1)
+
+    # Should we force the value ?
+    for  propname in "${forced[@]}"
+    do
+        # Should protect . in the $property
+        nameRegExp=$(echo $name | sed 's/\./\\\./g')
+        echo $name | grep -q "^$propname" && echo "FORCING : $property" >> $LOG && sed -i "s/^$nameRegExp=.*/$property/g" $cfgOut
+    done
+
+    # Should we skip this value ?
+    for  propname in "${skipped[@]}"
+    do
+        echo $name | grep -q "^$propname" && echo "SKIPPING : $name" >> $LOG && skip=1 && break
+    done
+    [ $skip -eq 1 ] && continue
+
+    # if the property doesn't exist, add it
+    grep -q "$name" $cfgOut || ( echo "ADDING to $cfgOut : $property" >> $LOG ; echo "$property" >> $cfgOut )
+done
+
+# Update is done, set the right version number
+cp $rbxVersion $curVersion
+
 # END


nadenislamarre:
@substring
- can you do an external script, or at least a function ?
- please do checks like :
  if ! sed ...

substring:
Ok for a function
Why use sed for checks ?

nadenislamarre:
if sed fails, at least log an error

digitalLumberjack:
```
# Shall we run the recalbox.conf update process ?
currentVersion=$(cat $rbxVersion)
lastConfUpdate=$(cat $curVersion 2>/dev/null)

containsElement () {
  local e
  for e in "${@:2}"; do [[ "$e" == "$1"* ]] && return 0; done
  return 1
}

containsKey () {
  [[ "$1" != "" ]] && [ ${userValues[$1]+abc} ] >> /dev/null && return 0
  return 1
}

[ "$currentVersion" == "$lastConfUpdate" ] && echo "recalbox.conf already up-to-date" >> $LOG && exit 0

echo "UPDATE : recalbox.conf" >> $LOG
savefile=${cfgOut}-pre-$(cat $rbxVersion)
mv $cfgOut $savefile

IFS=$'\n'
declare -A userValues
for property in `cat $savefile | grep -v '^[#;]\|^$'`;do
  userValues["`echo $property | cut -d '=' -f 1`"]=`echo "$property" | cut -d '=' -f 2`
done
for property in `cat $cfgIn`;do
      name=$(echo $property | cut -d '=' -f 1)
      realname=$(echo $name | sed 's/^;//g')
      # Should we force the value ?
      if containsElement "$name" $forced; then
        # Should protect . in the $property
        nameRegExp=$(echo $realname | sed 's/\./\\\./g')
        echo "FORCING : $property" && sed -i "s/^$nameRegExp=.*/$property/g" $cfgOut
        unset userValues["$realname"]
      elif containsElement "$name" $skipped; then
        # Should we skip this value ?
        echo "SKIPPING : $name" 
        unset userValues["$realname"]
      else
        if containsKey "$realname"; then
          echo "ADDING user defined to $cfgOut : $property"; echo "$realname=${userValues[$realname]}" >> $cfgOut
          unset userValues["$realname"]
        else
          echo "ADDING default to $cfgOut : $property"; echo "$property" >> $cfgOut
        fi
      fi
done
echo remaining options
echo "${!userValues[@]}"
# Update is done, set the right version number
cp $rbxVersion $curVersion
```

digitalLumberjack:
1 - get defined variables in an array
2 - clear user recalbox.conf
3 - read share_init recalbox.conf
4 - for each line : 
- if forced, just use the share_init value
- if skipped, do not write value
- else 
  - use the user defined value
  - or default value

substring:
Not sure it works right : values in the skipped were overridden but the share_init values (i intended skipped to work as "don't compare and keep user's values", whereas as now it works upside down), and the whole share_init recalbox.conf was appended to the end of the updgraded recalbox.conf
- you can't `mv $cfgOut $savefile` if $savefile already exists

substring:
Here is a slightly better version :

``` bash
# Shall we run the recalbox.conf update process ?
diff -N "$curVersion" "$rbxVersion" > /dev/null 2>&1 && echo "recalbox.conf already up-to-date" >> $LOG && exit 0

containsElement () {
  local e
  for e in "${@:2}"; do [[ "$e" == "$1"* ]] && return 0; done
  return 1
}

containsKey () {
  [[ "$1" != "" ]] && [ ${userValues[$1]+abc} ] >> /dev/null && return 0
  return 1
}

echo "UPDATE : recalbox.conf" >> $LOG
savefile=${cfgOut}-pre-$(cat $updVersion)
rm $savefile 2>/dev/null
mv $cfgOut $savefile

declare -A userValues
while read property ; do
  name=`echo $property | cut -d '=' -f 1`
  nameRegExp=$(echo $name | sed 's/\./\\\./g')
  value=`echo $property | sed "s/$nameRegExp=//g"`
  userValues["$name"]="$value"
done < <(grep -v '^[#;]\|^$' "$savefile")

# Add each line of the share_init recalbox.conf, and decide what happens if a property is forced, custom or 
while read property ; do
  name=$(echo $property | cut -d '=' -f 1)
  realname=$(echo $name | sed 's/^;//g')
  # Echo empty lines for \n
  if [[ $property == "" ]]; then
    echo "" >> $cfgOut
  # Should we force the value ?
  elif containsElement "$name" $forced; then
    # Should protect . in the $property
    #nameRegExp=$(echo $realname | sed 's/\./\\\./g')
    #echo "FORCING : $property" && sed -i "s/^$nameRegExp=.*/$property/g" $cfgOut
    echo "FORCING : $property" ; echo $property >> $cfgOut
    unset userValues["$realname"]
  elif containsElement "$name" $skipped; then
    # Should we skip this value ?
    echo "SKIPPING : $name" 
    unset userValues["$realname"]
  else
    if containsKey "$realname"; then
      echo "ADDING user defined to $cfgOut : $property"; echo "$realname=${userValues[$realname]}" >> $cfgOut
      unset userValues["$realname"]
    else
      # need to protect
      echo "ADDING default to $cfgOut : $property"; echo $property >> $cfgOut
    fi
  fi
done < $cfgIn
echo "Remaining options :"
echo "${!userValues[@]}"
# Update is done, set the right version number
cp $rbxVersion $curVersion
```

It has 2 defaults :
- it breaks the only escaped caracter in `## Escape your special chars (# ; $) with a backslash : $ => \$` (the backslash disappears)
- if a property (ie name=value) exists both as commented and non commented, it will be set to the 4.1 version. That's the case with `snes.core=snes9x_next` (it's fully commented on my test setup, and was defaulted to snes9x_next once upgraded)

the 1st point is not important. But for the second one, shall we bother ?

We still need to add the remaining values. Nothing difficult with that

substring:
OMG just thought about 1 big problem : if the wifi password had protected characters, they will be unprotected by the update script. Unless the security feature that encrypts passwords is run before the recalbox.conf update script
Or we should add in the script a way to protect the chars upon reading
Edit : solved => `read -r`

substring:
closing this one, will open a new PR

PR #484, opened 2016-09-21
Upgrade recalbox.conf : 2nd attempt

substring:

Fixes #XXXX

Changes :
This version is now a standalone function in a new .sh
Now it can :
- copy over protected values (like in the wifi password)
- append user defined properties that don't exist by default
- logs in recalbox.log
- property values can have a = inside (like for gamecon), still works
  Only drawback : takes about 7s on my pi3. Well, that's once every upgrade, people on the unstable branch will have to cope with it

Please comment and review

### DIFF
--- a/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
+++ b/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
@@ -47,7 +47,6 @@ function doRbxConfUpgrade {
   # so we can report them back in the new recalbox.conf
   declare -A userValues
   while read -r property ; do
-    # [[ $property =~ ^[#\;]{0,}$ ]] && continue
     name=`echo $property | cut -d '=' -f 1`
     nameRegExp=$(echo $name | sed 's/\./\\\./g')
     value=`echo -E $property | sed "s/$nameRegExp=//g"`
@@ -79,9 +78,6 @@ function doRbxConfUpgrade {
       fi
     fi
   done < $cfgIn
-
-  echo "Remaining options :"
-  echo "${!userValues[@]}"
 
   if [[ ${#userValues[@]} -gt 0 ]] ; then
     echo -e "\n## User custom properties" >> $cfgOut
--- a/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
+++ b/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
@@ -34,7 +34,7 @@ function doRbxConfUpgrade {
   curVersion=$_SHARE/system/logs/lastrecalbox.conf.update
   updVersion=$_RBX/recalbox.updateversion
   forced=(controllers.ps3.driver)
-  savefile=${cfgOut}-pre-$(cat $updVersion)
+  savefile=${cfgOut}-pre-$(cat $rbxVersion | sed "s+[/ :]++g")
 
   # Check if an update is necessary
   diff -N "$curVersion" "$rbxVersion" > /dev/null 2>&1 && recallog "recalbox.conf already up-to-date" && return 0
--- a/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
+++ b/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+
+#
+## This is a knid of header file that can be used to declare variables and functions
+## That could turn useful in any recalbox script
+#
+
+#
+## Variables
+#
+
+_RBX=/recalbox
+_SHAREINIT=$_RBX/share_init
+_SHARE=$_RBX/share
+
+#
+## Functions
+#
+
+# Checks if $1 exists in the array passed for $2
+function containsElement {
+  # local e
+  # for e in "${@:2}"; do [[ "$e" == "$1" ]] && return 0; done
+  [[ "${@:2}" =~ "$1" ]] && return 0
+  return 1
+}
+
+# Upgrade the recalbox.conf if necessary
+function doRbxConfUpgrade {
+  # Update recalbox.conf
+  cfgIn=$_SHAREINIT/system/recalbox.conf
+  cfgOut=$_SHARE/system/recalbox.conf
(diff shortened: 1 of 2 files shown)


board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
```diff
@@ -0,0 +1,95 @@
...
+# Upgrade the recalbox.conf if necessary
+function doRbxConfUpgrade {
+  # Update recalbox.conf
+  cfgIn=$_SHAREINIT/system/recalbox.conf
+  cfgOut=$_SHARE/system/recalbox.conf
+  rbxVersion=$_RBX/recalbox.version
+  curVersion=$_SHARE/system/logs/lastrecalbox.conf.update
+  updVersion=$_RBX/recalbox.updateversion
+  forced=(controllers.ps3.driver)
+  savefile=${cfgOut}-pre-$(cat $updVersion)
```
digitalLumberjack: savefile=${cfgOut}-pre-$(cat $rbxVersion)

board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
```diff
@@ -0,0 +1,95 @@
...
+}
+
+# Upgrade the recalbox.conf if necessary
+function doRbxConfUpgrade {
+  # Update recalbox.conf
+  cfgIn=$_SHAREINIT/system/recalbox.conf
+  cfgOut=$_SHARE/system/recalbox.conf
+  rbxVersion=$_RBX/recalbox.version
+  curVersion=$_SHARE/system/logs/lastrecalbox.conf.update
+  updVersion=$_RBX/recalbox.updateversion
```
digitalLumberjack: This file does not exist.
We already are on the new version as the update has already be done.
We can remove this file.

digitalLumberjack (changes requested):
Nearly perfect :)

board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
substring: would need some string mnipulation as it has afull datetime in it ( some / and : )
In that case, the curVersion var is not necessary either

digitalLumberjack: right

substring:
@digitalLumberjack Changes made as requested. I preferred keeping the `~/logs/logs/lastrecalbox.conf.update` file and the `$curVersion` if the users decided to remove recalbox.conf backups. At least, that way, we manage in a different place the recalbox.conf version

board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
```diff
@@ -0,0 +1,95 @@
...
+  # Add each line of the share_init recalbox.conf, and decide what happens if a property is forced, custom or 
+  while read -r property ; do
+    name=$(echo $property | cut -d '=' -f 1)
+    realname=$(echo $name | sed 's/^;//g')
+
+    # Echo empty lines for \n
+    if [[ $property == "" ]]; then
+      echo "" >> $cfgOut
+    # Should we force the value ?
+    elif containsElement "$name" $forced; then
```
digitalLumberjack: elif containsElement "$name" "${forced[@]}"; then
Last effort :)

digitalLumberjack (changes requested):
Just a little change

substring:
@digitalLumberjack Requested changes were made + slight little improvements. See the commit for details

digitalLumberjack (approved):
LGTM

substring:
@nadenislamarre your review would be appreciated pal :)

nadenislamarre:
sorry, i'm not fan of your shell style as already told.
i've written quickly a similar one in shell, so you can inspire it to update yours, as you prefer.
you can run my example with : ./updateRecalboxConf.sh recalbox.conf.old recalbox.conf

[updaterecalbox.tar.gz](https://github.com/recalbox/recalbox-buildroot/files/509340/updaterecalbox.tar.gz)

substring:
I like :
- the easier steps (copy the shareinit file to a temp file, look for keys from the share and rewrite them in the temp file)
- much faster
- error handling
- less arrays (which is a consequence of the 1st point)

I don't like :
- you expect some indentation with `[ ]*` as well as some spaces around the `=`.
- you're not using an array for forced values, it would make things easier to maintain and parse
- I don't see the aim of the LAST_KEY. If it's to make a difference in versions, there are much better options
- doesn't replace inline properties that were uncommented, it just writes them at the end of the file (ex : set some wifi SSID and PASS, they are appended instead of replaced)
- missing some logging (ok, that's a quick and dirty version)
- paranoid mode : exits on any error
- it doesn't even make coffe

@digitalLumberjack shall I try to make the best of the 2 scripts ?

nadenislamarre:
> > you expect some indentation with [ ]\* as well as some spaces around the =.
> > i've not understood. Feel free to put the regular expression that works.
> > 
> > you're not using an array for forced values, it would make things easier to maintain and parse
> > keep it simple.
> > why make something complicate for something that will probably never evolve ?
> > more over, in case it evolves, we can change it. Note that if it evolves in a sense that regular expression make more sens, like ps3.*, it can be easier to maintain.
> > I prefer to make things more complex when they really happend.
> > 
> >    I don't see the aim of the LAST_KEY. If it's to make a difference in versions, there are much better options
> > you didn't want my revision property ;-( so it's just a quick way to check that the file is up to date. When you add values into recalbox.conf, you've to update this value : it's the drawback compared to the revision property.
> > 
> >    doesn't replace inline properties that were uncommented, it just writes them at the end of the file (ex : set some wifi SSID and PASS, they are appended instead of replaced)
> > in fact, i've just tested the file i put in example. So, possibly i don't handle some cases.
> > 
> >    paranoid mode : exits on any error
> > if you check what can cause the error, it's only disk errors or / in values that makes sed fails (because i've not escape them).
> > I don't find it paranoid, just safe.
> > 
> >    it doesn't even make coffe
> > yes, this was the main target of this script : show what a non java style shell script looks like ;-)

PR #490, opened 2016-09-27
x86 : add Ubuntu's blacklists

substring:

Fixes #XXXX

Changes :
- add modprobe blacklists

as requested by @nadenislamarre
Helps some boards to work, including my PC


substring:
i just put all of the blacklist files on my key. which is strictly that PR. All the blacklist files from my ubuntu that were put non my key are in this PR. Would be way too hard to sort them out

nadenislamarre:
i've found an ubuntu more recent than my pc at work and the file alsa.conf
doesn't exist anymore.
However, it was a server version and the files seems differents than yours.
my remark was because i checked on my pc with an old ubuntu.

so, i think your pr is good, thanks.

digitalLumberjack:
Let's go with this one :)

PR #496, opened 2016-09-30
apple2: fix for x86* and xu4

nadenislamarre:
it should be fixed.
however i still don't like the .mk while it refers to /recalbox/share*


digitalLumberjack:
nice

PR #510, opened 2016-10-17
es: fix the patch

nadenislamarre:


substring:
why have a patch when we have our own repo for ES ???

nadenislamarre:
because this patch is dependant of an sdl patch.
making it as a patch lets es work on a development machine without patching
sdl.

substring:
Ah ok ! you're really trying to mess up my brain :smile: 

You could consider the `<pkg>_OVERRIDE_SRCDIR` directive when you need to test ES on recalbox with your local source :)

PR #536, opened 2016-10-29
Add libx264 to record through ffmpeg in retroarch

substring:
- [x] You added the changes in CHANGELOG.md

Fixes https://github.com/recalbox/recalbox-os/issues/447

Changes :
- add the libx264 ecoder to ffmpeg

We can now record in retroarch with this codec (which is the best suited for local recording or streaming). The Pi H264 hardware encoder in ffmpeg (h264_omx) requires ffmpeg3, but my tests so far showed that videos quality was rather bad. We'll see that with the next merge from buildroot.

Streaming to twitch does work, but has to be well setup to minimize the impact on CPU. I've made a few live sessions that some people could watch. I may add a cfg file for twitch if asked


digitalLumberjack:
Nice :)

PR #548, opened 2016-11-07
kodi: enable omx player by default

nadenislamarre:


substring:
Yo !
Are you sure omxplayer is available on platforms other than Pi ?

nadenislamarre:
it's not. but it's compiled without omx. the option will have no effect.
i prefer that to have to manage x kodi xml files for the moment.
it means : enable it if available.

digitalLumberjack:
alright

PR #549, opened 2016-11-09
dosbox: fix x86_64 compilation

nadenislamarre:
Existing patch :
Migrate dosbox to SDL2
See http://www.vogons.org/viewtopic.php?t=34770
Signed-off-by: Laurent Merckx <laurent-merckx@skynet.be>

Change :
=> I've put the sdl checks in configure.ac at the end of the file
while settings *FLAGS at the begining of the configure broke the basic configure steps (like checking char size returning 0 instead of 1 because of libSDL2*.so not found (i don't known why however while -L was specified, i've not found the deep deep cause)).
i've compiled rpi2, rpi3, c2, x86_64 to confirm it still compiles on all platforms.

### DIFF
--- a/package/dosbox/dosbox-001-sdl2.patch
+++ b/package/dosbox/dosbox-001-sdl2.patch
@@ -1,9 +1,7 @@
-Migrate dosbox to SDL2
-See http://www.vogons.org/viewtopic.php?t=34770
-
-Signed-off-by: Laurent Merckx <laurent-merckx@skynet.be>
---- dosbox/acinclude.m4	(revision 3924)
-+++ dosbox.n/acinclude.m4	(working copy)
+diff --git a/acinclude.m4 b/acinclude.m4
+index 936496e..1dbc45b 100644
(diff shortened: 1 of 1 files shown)


ian57:
Yep, saw that with substring yesterday... We did not known why I had to install sdl2-dev on the host X86_64 computer to be able to compile the X86_64 recalbox dosbox . Thanks for the patch ;)

nadenislamarre:
you dont need it anymore

Le jeu. 10 nov. 2016 à 05:59, Yann Morere notifications@github.com a
écrit :

nadenislamarre:
on the configure.ac the line causing the pb is
cppflzgs=-L... -lSDL2 too soon which makes basic check to not compile as if
-L where not used

nadenislamarre:
when it fails, grep libSDL congfig.log : libSDL.so file not found

digitalLumberjack:
I can only trust you on this one :P

PR #562, opened 2016-11-16
Rb 4.1.x nadenislamarre libretro glupen64

nadenislamarre:
make it compile for x86 x86_64 odroids c2 and xu4

note1 : x86 compilation is not tested while i've no space to keep all the archs on my computer.
note2 : compile doesn't mean it works. but i really trust it will. testers are welcome.


paradadf:
It would be good to have the same button mapping as mupen64plus.

nadenislamarre:
is it different ? did you do a test ?

rockaddicted:
yes it is different.

-mupen standalone version : 

| N64 pad | ES pad (snes) |
| :-: | :-: |
| A | B |
| B | Y |

-mupen libretro version (glupen64) : 

| N64 pad | ES pad (snes) |
| :-: | :-: |
| A | A |
| B | B |

digitalLumberjack:
Open an issue for the mapping if necessary.

PR #564, opened 2016-11-17
locale: fix one compilation error

nadenislamarre:


digitalLumberjack:
Oups sorry for that ;)

PR #567, opened 2016-11-27
Update fba_libretro gamelist.txt

paradadf:
Just a little update taken from:
https://github.com/libretro/fbalpha/blob/master/gamelist.txt


digitalLumberjack:

PR #577, opened 2016-12-19
Initial steam-controller support

bpollman:
To enabled service, set "controllers.steam.enabled=1"

Resolves Issue: recalbox-os/#466

Changes :
- added package for python-libusb1
- added package for https://github.com/ynsta/steamcontroller 
- added steam controller service


digitalLumberjack:

SigmaticMinor:
Ooo!!! Looking forward to testing this on mine, can anyone let me know how I can put a build with this onto my pi?

digitalLumberjack:
Nice. 
Can you provide the configs files too ? (edit manually or make savedefconfig on each configuration)

bpollman:
I updated all the configs except recalbox-arcade_defconfig since it looked like a stripped down release.  I expect it to build fine everywhere but I haven't personally tried it on anything except rpi1/rpi3

bpollman:
@SigmaticMinor You can do a build based on the steam-controller branch and install as normal and it should work. Or you can wait until it gets pulled into the master branch and build it then.  The controller is quite twitchy. But I guess thats by design.

substring:
Woot a nice contribution ! Welcome on board !

adding `controllers.steam.enabled` to every recalbox.conf with a comment would be awesome, so that people editing the file see the option right away. Now the question would be : shall will set it to 1 as default (if it's harmless) ?

bpollman:
@substring Thanks!

Enabling it by default is probably harmless but the service calls `sc-xbox.py start` and it takes several seconds to complete. May want to leave it disabled by default for now. 

Wasn't sure where the right spot for the steam support was in the recalbox.conf, D4?

```
# ------------ D4 - Steam Controllers ------------ #
# Enable support for Steam controllers (0,1)
controllers.steam.enabled=0
```

substring:
Well, using ampersands in the right place in the the Sxx init scripts should make this harmless :)

One last thing : edit the es_input.cfg to add the pad definition, so that it's configured ootb

bpollman:
Very true on the ampersands! But I'll take a look at cpu usage when it starts to make sure it doesnt have a significant impact on boot time. 

as for the es_input.cfg, Emulation Station is currently auto detecting it as a XBOX 360 Gamepad.  Is that sufficient or is there more to it? I'm still feeling my way around how ES works.

substring:
ES uses the oad name + its SDL2 GUID. It's supposed to be based on the vendor id, product id and revision id, something like that. If your controller was working straight once the build was ok, then no need to add anything. If you had to map the controller inside es, then you should add its definition. The file you should check is `~/.emulationstation/.es_input.cfg`. If you configured your steam controller, it's most probably at the end of the file

digitalLumberjack:
@bpollman hi :) did you have time to check if the controller must be configured ?

bpollman:
Hey. Not yet, will try to do a test today. My hardware has been in heavy use by friends and family over the holidays.

bpollman:
@digitalLumberjack @substring ampersands have been added and it looks like the controller 'just works'. I reset my es_input.cfg and the controller worked without any configuration needed.

digitalLumberjack:
great thanks

charlesmarried:
Hi, Im new here. I have 2 steam controller and I want to know if the Steam Controller works with Recalbox nowadays. Thank you!

PR #586, opened 2016-12-28
Set gpu_mem_1024=256 (RPi2 & RPi3)

paradadf:

Fixes #XXXX

Changes :
- Allocate less memory to the gpu to have more ram available.
Fixes problems with fba games with samples (fireshrk, ghox, vimana) not launching at all because of memory constraints.
EmulationStation using the official recalbox themes doesn't need so much gpu as before. This has been tested with 43 available systems without problems.
- I have tested n64 games and couldn't notice any performance drawback with this change.


digitalLumberjack:

digitalLumberjack:
Did you check with dreamcast and ppsspp ?

paradadf:
I can't test those because I only have a RPi2.

digitalLumberjack:
okay we will test before merging :)

digitalLumberjack:
The second issue i think about is for emulationstation themes that uses png for background. 512mo was able to store ~25 background images not more. So we have to refactor es or this PR will break many recalboxes :)

PR #593, opened 2017-01-01
libcapsimage: fix the link on the target (fix atarist)

nadenislamarre:


substring:
Are you linking the `/usr/lib/libcapsimage.so.4.2` from your PC ???

digitalLumberjack:
the links keeps the absolute file path so it will work as intended while running on the target.

PR #594, opened 2017-01-01
Fix build

digitalLumberjack:
Removed wiring pi
removed steam controller from odroid (kernel incompatibilites)
fixed x86 image generation

@rockaddicted @substring we should create issues for thing we might reactivate.


substring:
@digitalLumberjack done https://github.com/recalbox/recalbox-os/issues/1159

PR #615, opened 2017-01-14
added update of boot after upgrading recalbox version

digitalLumberjack:
Update /boot/config.txt with new values for recalbox 4.1 after an update.

Not my proudest one @rockaddicted @substring if you have better ideas i take them.

### DIFF
--- a/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
+++ b/board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
@@ -25,6 +25,13 @@ function containsElement {
   return 1
 }
 
+function shouldUpdate {
+  rbxVersion=$_RBX/recalbox.version
+  curVersion=$_SHARE/system/logs/lastrecalbox.conf.update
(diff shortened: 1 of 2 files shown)


board/recalbox/fsoverlay/recalbox/scripts/recalbox-includes.sh
```diff
@@ -68,3 +75,8 @@ function doRbxConfUpgrade {
   cp "$rbxVersion" "$curVersion" || { recallog -e "ERROR : Couldn't set the new recalbox.conf version" ; return 1 ; }
   recallog "UPDATE done !"
 }
+
+
+function updateBoot {
+  sed -i '/^dtparam=/{h;s/=.*/=audio=on/};${x;/^$/{s//dtparam=audio=on/;H};x}' /boot/config.txt
```
substring, line 81: could be worth adding some recallog + return value, should we ever fail an update
Well, no afterall, because the share is not mounted at S04

substring (changes requested):

substring (approved):

PR #660, opened 2017-01-29
add upnp and smb protocol support for kodi

david-barbion:
Kodi in RB 4.1.X does not allow to connect to UPNP nor SAMBA servers.
This PR adds Samba4 package and enables uPnp and Samba in Kodi.


digitalLumberjack:
Hi, thanks for the contribution :)

we now develop on the master branch, where we add the support for odroid board and x86/64.

Could you rework your pr on the branch, editing also the odroid and x86 defconfigs ?

PR #661, opened 2017-01-29
set kodi to be executed as kodi user

david-barbion:
Currently, Kodi is executed as root. This PR add the ability to mount NFS share without "no_root_squash" options. This allows a fine grained permission setting on the NFS server.


digitalLumberjack:
I see a problem here, we distribute kodi config files here : https://github.com/recalbox/recalbox-buildroot/tree/master/board/recalbox/fsoverlay/recalbox/share_init/system/.kodi

They belong to root. So on a fat 32 fs it would be ok, but not on ext...

digitalLumberjack:
@substring

substring:
- rights problems may not occur on FAT32 if they are mounted as 777. But for an ext4 mount, where files would prolly belong to root ... so yes
- why only pi and not other archs ?
- this can be solved server side

PR #667, opened 2017-02-22
Use the sound settings from the configuration file

redbug26:
Changes :
- Use the sound settings from the configuration file


substring:
I love it !! I'll review that tomorrow :)

substring:
At S02, we don't know yet which is the device that will be mounted on /recalbox/share, so users with a USB key or CIFS won't see their volume adjusted according to their setting. Sadly, that's something we can't solve easily,. Which means @redbug26 you came with the right solution !

Changing the audio output device is a great idea !

substring (dismissed):

rockaddicted:
Hi and thank you for your PR.
Could be awesome to also check this option `audio.bgmusic=1` (or a new one like `audio.splashmusic=1`) to totally disable sound of splashscreen.

What do you think about that ?

nadenislamarre:
at s02, you've no access to /recalbox/share/system/recalbox.conf too even when using internal devices, not only usb keys.

redbug26:
True...
What do you think about a script that modify the S02splash file when you shutdown the raspberry ?
It could be easily done too.

substring:
You do not edit an init script ! Updates would erase changes !
2 solutions :
- store volume in /boot
- store volume in /

/boot looks like the best choice, in recalbox-boot.conf for ex. @digitalLumberjack any opinion ?

digitalLumberjack:
Yep @substring recalbox-boot.conf seems to be the best solution but /boot is not mounted when S02 runs.
And we won't modify /.

redbug26:
Thus ? 

On ne fait rien et on laisse comme ça ? :-)

digitalLumberjack:
No we can try to move the boot mount before the splashscreen :)

substring:
Problem : S04boot won't be erased (yet) on upgrade to 4.1 ...

substring:
@redbug26 have you read `/etc/init.d/rcK`? ;)
And we "prefer" using python for reading the recalbox.conf, that's how it's done in all other scripts

board/recalbox/fsoverlay/etc/init.d/K99cache
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
substring: Our init uses the stop parameter, not K-prefixed scripts

redbug26:
It was just some new tests on my repo :-)
I'll look for the stop parameters and the python things tonight.
Thanks for your return

PR #674, opened 2017-04-25
PPSSPP v1.4

Ogekuri:
Changes :
Upgrade PPSSPP to version 1.4
Fix Kodi Youtube plugin to version 5.3.8
Fix Disable_mouse patch for PPSSPP v1.4
Fix gamecontrollerdb_patch.patch for PPSSPP v1.4
Fix makelist.patch for PPSSPP v1.4
Removed libpng.patch


substring:
Hi !

Thanks for your contribution​ !

Can you explain why you removed the libpng patch ? As far as i remember it's mandatory for the Odroid C2

Ogekuri:
Honestly I've only tested the Raspberry Pi 3 build; for Pi3 a libpng v1.7 seems statically linked so the resulting binary not load the system's libraries. The system still use the version 1.6 of libpng. I've only tested it with few games (Tekken 6, LEGO Batman, Crash and Cube).. but it's seem works -at least- as before.

PR #678, opened 2018-07-26
Update recalbox-encode.sh

jalbersol:
Prevent line breaks in long wifi passwords with more than 30 characters.

Changes :
- Long wifi passwords with more than 30 characters will not longer cut.


substring:
Hi !
Thank you for this MR. I'll review it once back from holidays

PR #681, opened 2019-11-14
Unable to change batocera ROM folder

AradielSabathius:
Hello.
I need to know how to change the roms destination folder.
Actually I boot batocera/recalbox from an 8 GB compact flash card with sata adaptor.
Also I have a 250 GB sata hard disk to store roms.
Everything works properly. No problems working by the moment.
Since 8 GB is not enough to store all games, ¿How can I change the ROMS folder to the 250 GB hard disk?
Thanks.


MikaXII:
Hello @AradielSabathius,
Please, a little more attention ^^, you're in the wrong repository => https://github.com/batocera-linux/batocera.linux

